PR #8, opened 2015-02-02
Bug/split

nathanielguy:
split.xcmsSet in master doesn't work (Peaks are lost, wrong phenodata is copied).  Here is a fix that works as expected.

First pull request and I don't know if I followed all your conventions so someone should check this over.

### DIFF
--- a/R/xcmsSet.R
+++ b/R/xcmsSet.R
@@ -225,12 +225,19 @@ split.xcmsSet <- function(xs, f, drop = TRUE, ...) {
   if (any(! f %in% xs@phenoData[,"class"])) {
     stop("Non-existant class specified.")
     }
+
+  split.samps = lapply(f, function(x) {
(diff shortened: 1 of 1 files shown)


nathanielguy:
I was mistaken about the form of f and as such my pull request is not a drop in replacement and should not be merged.

I need to revise this pull request or replace it with another.  This issue is:  Output of split.xcmsSet is nonsensical when length(f) < nrow(x@phenodata).

I will submit a second pull request which raises an error when length(f) != nrow(x@phenodata).

Sorry for the clutter.

nathanielguy:
Closing this request.

sneumann:
No worries, thanks for looking into this. 
Can you include a test case, ideally a unit test that your patch fixes ?
Actually, length(f) < nrow(x@phenodata) does make sense, 
if e.g. every n-th sample is a QC sample you want to remove.

PR #10, opened 2015-03-11
Fixed split.xcmsSet in the case of length(f) != nrow(xcmsSet@phenodata)

nathanielguy:
I rewrote my pull request to implement the fix in one commit and in the fewest number of changes possible.  I thought this would be easier to verify.

### DIFF
--- a/R/xcmsSet.R
+++ b/R/xcmsSet.R
@@ -236,28 +236,29 @@ split.xcmsSet <- function(x, f, drop = TRUE, ...) {
 
     lcsets <- vector("list", length(levels(f)))
     names(lcsets) <- levels(f)
-
+    
     for (i in unique(sampidx)) {
+        samptrans = which(sampidx == i)
+        samptrans = samptrans[samptrans <= nrow(x@phenoData)]
+        
+        if (length(samptrans) < 1) next
+        
         lcsets[[i]] <- new("xcmsSet")
-
-        samptrans <- numeric(length(f))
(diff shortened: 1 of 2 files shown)


nathanielguy:
There is also functionality on the parent branch of this which retains group information after splitting.  Its not ready for the prime time but its functional for my uses.  If its needed we can work on incorporating it.

sneumann:
Hi Nathaniel, 
sorry for taking so long. it's great that you also included the corresponding unitTest.
What exactly is the semantics of your change ? In my opinion if f != nrow(phenodata)
the function call would be invalid, so I'd rather stop() with a good error message.

Regarding retaining the group information, this must give the same result as an actual group(),
and having fewer samples will likely lead to different density estimations, thus different groups.
Obtaining the identical result with group() also requires honouring the original minfrac,
which is not recorded in the xcmsSet. Since the group()ing is pretty fast, I doubt getting this right 
is worth the effort. I would accept a patch if there is a rigorous unitTest included. 

Yours,
Steffen

nathanielguy:
Hi Steffen,

The change makes split.xcmsSet() function in the same way base split() does. (Which I believe was the desired behavior based on the drop= variable and documentation.)

Eg. split(c(1,2,1,2), c("a","b")) returns a list with 1's and 2's separated.

Also, in our previous conversation on the closed pull "Bug/split #8" you hinted at a more appropriate solution than simply raising an error which encouraged me to write this fix.

Nate

PR #11, opened 2015-04-21
Some minor tweaks: using parallel package for parallel processing if MPI or snow is not available

jotsetung:
+ Added $ method to access phenoData columns in a eSet/ExpressionSetlike manner.
+ Allow to use the "parallel" package for parallel processing of the functions xcmsSet and fillPeaks.chrom.
+ Added relevant documentation entries.

### DIFF
--- a/R/xcmsSet.R
+++ b/R/xcmsSet.R
@@ -127,7 +127,7 @@ xcmsSet <- function(files = NULL, snames = NULL, sclass = NULL, phenoData = NULL
             } else if (parMode == "parallel"){
                 ## using the mclapply to run the code in parallel on the multiple
                 ## CPUs of the actual host/computer.
-                res <- mclapply(argList, findPeaksPar)
+                res <- parallel::mclapply(argList, findPeaksPar)
          } else {
               ## serial mode
               res <- lapply(argList, findPeaksPar)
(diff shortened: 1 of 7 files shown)


jotsetung:
Added now also the possibility to submit color definitions other then numeric (e.g. in html notion) to some of the plotting functions.

sneumann:
Hi Johannes, thanks for the patches. 
Can I ask you to address this warning:
````
* checking dependencies in R code ... WARNING
'library' or 'require' call not declared from: ‘parallel’
'library' or 'require' call to ‘parallel’ in package code.
  Please use :: or requireNamespace() instead.
````
Thanks in advance, yours, Steffen

File: R/xcmsSet.R [hunk 1]
```diff
@@ -1036,11 +1032,11 @@ setMethod("retcor.peakgroups", "xcmsSet", function(object, missing = 1, extra =
 
         screen(2)
         par(mar = c(5.1, 4.1, 0, 2), yaxt = "n")
-        allden <- density(peakmat[,"rt"], bw = diff(rtrange)/200, from = rtrange[1], to = rtrange[2])[c("x","y")]
+        allden <- density(peakmat[,"rt"], bw = diff(rtrange)/200, from = rtrange[1], to = rtrange[2], na.rm=TRUE)[c("x","y")]
```
sneumann: This looks like it is not only color cosmetics, but actually a fix in the density estimation. In what cases did you get NA's ? 
I'd like to add that to the NEWS file

jotsetung: hi Steffen,
concerning the first note, I'll address that. To the second, I inserted that code when I was looking for the reason I got the error, which was however related to the col parameter. I will remove that again and submit the code again.

sneumann:
Wait, why removing if it fixes something ? I just want a note "remove NA values from the rector(method=loess) correction if XXX" to document the change.

jotsetung:
Well, I'm not sure whether that fixes something. I didn't get an error based on that, the error came from the max(col) if the colors are not numeric... I was just looking first in the wrong place and forgot to remove that na.rm=TRUE again.

jotsetung:
Now the Note related to the require("parallel") should be fixed (using requireNamespace and parallel::mclapply instead). Also, I removed the two na.rm=TRUE that i added but which are not required.

PR #12, opened 2015-05-05
Some minor and larger changes

jotsetung:
dear Steffen,

this are the changes I made:
New features/changes:
+ levelplot for xcmsRaw and xcmsSet objects (similar to the image plots for xcmsRaw). The method for xcmsSet allows to highlight the identified peaks.
+ getXcmsRaw method for xcmsSet objects: reads the raw data, applies eventual corrections and returns it as an xcmsRaw object.
+ Added arguments "lty" and "add" to plotEIC method (plot EIC from multiple files to the same plot).
+ xcmsSet now allows argument phenoData to be an AnnotatedDataFrame.

The biggest change I made was in the method getEIC which I completely rewrote. I was puzzled by the fact that the function did only allow to calculate the EIC for a limited rt and m/z range (the original method did not return correct chromatograms for m/z ranges larger 100). My version allows to extract the ion chromatogram for the full rt and m/z range (i.e. the base peak chromatogram). I cross-checked the results between the new and the old function and they match. I also added an unit test to make sure the method returns what is expected.

Bug fixes:
+ xcmsSet getEIC method for water Lock mass corrected files for a subset of files did not evaluate whether the specified files were corrected.
+ Checking if phenoData in xcmsSet function is a data.frame or AnnotatedDataFrame and throw an error otherwise.

In addition I updated/elaborated a bit on the documentation.

I hope you find my changes useful.

cheers, jo


sneumann:
Dear Jo, this is excellent!

I have a few comments:
1. The levelplot stuff is great, and could go in as-is.
2. The phenoData stuff also sounds good, but I'd need to see
if that affects e.g. Risa, which can create phenoData from ISAtab files for you.
Once that is checked on one or two ISA files (faahKO contains some, 
and also https://github.com/sneumann/mtbls2
3. I am a bit hesitant about the new getEIC, so I'd like to have 
   both the old and the new one at least during the development cycle. 
   The reason is that CAMERA uses getEIC, and I'd like to compare 
   the behaviour between the two. One could hack an option() and a big if() 
   that switches depending on that option, rather than commenting out as it is currently.
4. the getXcmsRaw() is excellent and was missing a long time. 
  Do you think all cases are covered to retrieve the same xcmsRaw 
  that was used during e.g. findPeaks ? Things that modify/filter the xcmsRaw 
  are also the scanrange=c() or the polarity= 
  What really would be required is a slot in the xcmsSet that keeps track 
  of what parameters were used in the creation, so that these arguments 
  can be re-applied for getXcmsRaw(). These could also be shown in the summary()
  for an xcmsSet.

So I am happy to take 1. and 2. right away, 3. after that change and 4. if it would be 
possible to add the mentioned book-keeping infrastructure soon. Essentially, 
getXcmsRaw should allow to round-trip (only that currently we don't have 
an xcmsSet constructor that would take the xcmsRaw's from getXcmsRaw...)

I hope the comments are not too frustrating and rather encouraging to continue on it. 

Yours,
Steffen

jotsetung:
dear Steffen!
That's good news. I completely understand your concerns related to point 3). Actually, one alternative suggestion would be to use the method as-is for m/z ranges smaller 100 bins and my version for m/z ranges larger 100 bins. Would that be an alternative option you could accept?
Concerning 4) I absolutely agree. I did not consider e.g. the scanrange. I think of a way to include a @history slot in the xcmsSet object.

sneumann:
Hi, w.r.t. the getEIC() I would like to replace the old implementation if we see that it gives 
the same results for normal EICs and better for the big ones. 
Two implementations would be a nightmare to maintain in the long run. Just make sure the pull requests 
are more isolated. If you leave out the version bump in the DESCRIPTION and put the NEWS stuff into the commit message, we can even merge them in (almost) any order.

jotsetung:
OK, I've added the option BioC$xcms$getEIC.method that allows to switch between the original getEIC method (setting the option to "getEICOld") and my new implementation (option "getEICNew"). The getEIC method switches between the two methods based on this option. In runit.getEIC.R I have an example how to switch.
In addition, I've added some more slots to xcmsRaw and xcmsSet that allow to record the mslevel and scanrange settings. That then used in the getXcmsRaw to retrieve the data with exactly the same settings that were used with the xcmsSet function. For backwards compatibility I'm checking in the getter/setter methods whether the slots actually exist (e.g. if an older version xcmsSet object was saved and loaded).
I've also added some more methods so that methods and usability for xcmsSet and xcmsRaw is somewhat similar.
Regarding the polarity argument in the xcmsSet function, I was puzzled by the fact that the xcmsRaw function to read the raw data doesn't have such an argument... is there something missing?
Also, I would have a suggestion that might make future developments a little easier: putting all generics into the 00Generics.R file. I would do that if it's OK with you.

cheers, jo

hpbenton:
Johannes - the xcmsRaw reads the polarity from the file. That way each xcmsRaw object has the polarity of the file. The xcmsSet can then process either polarity if the switch is set. The getXcmsRaw method sounds nice.
cheers,
Paul

PR #16, opened 2015-05-15
Fixed a bug that resulted in a cryptic error message when no input files are available to the xcmsSet function.

jotsetung:
By chance I came across this problem and it took me a while to figure out that actually I misspelled the input file names... I moved the file checking up in the xcmsSet code so that a correct error message is shown instead.

cheers, jo

### DIFF
--- a/R/xcmsSet.R
+++ b/R/xcmsSet.R
@@ -334,7 +334,17 @@ setReplaceMethod("sampnames", "xcmsSet", function(object, value) {
 setGeneric("sampclass", function(object) standardGeneric("sampclass"))
 
 setMethod("sampclass", "xcmsSet", function(object) {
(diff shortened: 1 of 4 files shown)


jotsetung:
This pull request contains now also some changes to the `sampclass` and `sampclass<-` methods:
+ `sampclass` will return the content of a column named "class" from the `data.frame` in the phenoData slot, or, if not present, the interaction of all columns of that `data.frame` (the default *old* behavior).
+ The `sampclass<-` replaces the content of the column "class" in the phenoData and supports also a `data.frame` as input, in which case the interaction of all of its columns will be used to replace the content of the "class" column.
+ I've also added a RUnit test that checks that everything is working as expected.

cheers, jo

PR #18, opened 2015-05-21
Added [ method to xcmsSet and fixed a problem in split.xcmsSet

jotsetung:
+ I've implemented a [ method that allows to subset an xcmsSet.
+ Fixed a problem in split.xcmsSet that did not split the phenoData
properly.
+ Added some details to the documentation of xcmsSet-class.

hope you find the changes useful.
cheers, jo

### DIFF
--- a/R/xcmsSet.R
+++ b/R/xcmsSet.R
@@ -253,6 +253,18 @@ split.xcmsSet <- function(x, f, drop = TRUE, ...) {
     lcsets <- vector("list", length(levels(f)))
     names(lcsets) <- levels(f)
 
+    ## get the phenoData and all other parameters we want to pass
+    ## down to the splitted objects
+    pd <- phenoData(x)
+    dataCor <- x@dataCorrection
+    suppressWarnings(
+        msL <- mslevel(x)
+        )
+    suppressWarnings(
+        scanR <- scanrange(x)
+        )
+    pol <- x@polarity
+
     for (i in unique(sampidx)) {
         samptrans = which(sampidx == i)
         samptrans = samptrans[samptrans <= nrow(x@phenoData)]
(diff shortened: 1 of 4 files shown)


R/xcmsSet.R
```diff
@@ -253,6 +253,18 @@ split.xcmsSet <- function(x, f, drop = TRUE, ...) {
     lcsets <- vector("list", length(levels(f)))
     names(lcsets) <- levels(f)
 
+    ## get the phenoData and all other parameters we want to pass
+    ## down to the splitted objects
+    pd <- phenoData(x)
+    dataCor <- x@dataCorrection
+    suppressWarnings(
```
sneumann: Which warnings are expected here ? Please add to the comment, same three lines below.

R/xcmsSet.R
```diff
@@ -2296,3 +2323,157 @@ setMethod("profMethod", "xcmsSet", function(object) {
...
+  ## 5) filled
+  if(length(x@filled) > 0){
+      xsub@filled <- (1:nrow(keep.peaks))[rownames(keep.peaks) %in% as.character(x@filled)]
+  }
+  ## 6) dataCorrection
+  if(length(x@dataCorrection)>0)
+      xset@dataCorrection <- x@dataCorrection[j]
+  return(xsub)
+})
+## setMethod("[", "eSet", function(x, i, j, ..., drop = FALSE) {
```
sneumann: Once confident with the above, just remove this block in several weeks

man/xcmsSet-class.Rd
```diff
@@ -231,11 +239,21 @@
...
+      \code{signature("xcmsSet")}: divide the xcmsSet into a list of
+      xcmsSet objects depending on the provided factor. Note that only
+      peak data will be preserved, i.e. eventual peak grouping information
+      will be lost.
     }
     \item{\code{object$name}, \code{object$name<-value}}{
       Access and set \code{name} column in \code{phenoData}
     }
+    \item{\code{object[, i]}}{
+      Conducts subsetting of a \code{xcmsSet} instance. Only subsetting
```
sneumann: After subsetting the minfrac might be violated, right ?

R/xcmsSet.R
jotsetung: I've added the mslevel and scanrange slots in a previous commit, thus, they are not present in "older" xcmsSet objects (like the ones from the faahKO package). When the methods mslevel and scanrange are called on such objects a warning is generated stating that the slots are not available and returning the standard values for these slots.

R/xcmsSet.R
jotsetung: a yes, right. The commented out code can be removed straight away... sorry, forgot about that.

man/xcmsSet-class.Rd
jotsetung: yes, the subsetting of the groups might be problematic after all.

Issue #7, opened 2014-12-22
Problem with getEICs in xcms


sneumann:
Alan Smith reported:

We ran into a problem with the newer versions of xcms (1.38, 1.40, and 1.43 have been tested) after tyring to upgrade from version 1.26.1. Interestingly the newer versions of the getEIC function works fine on data xcmsSet objects generated under xcms version 1.26.1.  I just wanted to let you know that if this problem interests you and you would like real examples of our data I would be happy to provide it to you if the example below is not helpful.  Any advice on how to handle this issue will be greatly appreciated especially if it is a user error on my part.

We are experiencing problems generating EICs using the getEIC function with xcmsSet objects and group IDs when the "rt" argument is set to "cor".  Recently we upgraded xcms from version 1.26.1 to version 1.40 and began to experience the issue where all of the EICs generated by getEIC do not make sense when rt="cor".  The problem occurs with or without data that has been rt corrected when rt is set to cor.  The EICs look as expected when using rt = raw.  This problem occurs with every independent data set we have evaluated.    I have been able to reproduce the problem with the faahKO package [example code following session info].  I have attached a pdf from the example code below where page 1 and 3 correspond to the same feature.  We have rolled back to version 1.26.1, but would like to use the current version of xcms.  Any information on the origin of this problem and advice on how to solve this problem would be appreciated so that we can evaluate features with RT corrected data using the current builds of xcms and the getEIC fucntion.

sneumann:
Hi Alan, 
it would be a fantastic help if you could create a self-contained and standalone test here. 
The we could try to isolate the relevant commit between 1.26.1 and 1.44.x with "git bisect"
(see https://git-scm.com/book/en/v2/Git-Tools-Debugging-with-Git or https://robots.thoughtbot.com/git-bisect )
Yours,
Steffen

sneumann:
Alan Smith made some excellent progress:

Hi Steffen,

I picked this up again and found that the problems in EIC extraction are caused by the length of the vector of corrected retention times. The vector of corrected RTs corresponds to the length of the all of the scans in an xcmsRaw object in xcms version 1.26.1 and only the scan range used for peak detection in xcms version 1.46.0. 

[...]

PR #25, opened 2016-02-20
Fix for the Bug reported by Alan Smith in issue #7

jotsetung:
Dear Steffen,

I replaced the code in the getEIC to load the raw data by the getXcmsRaw method. That one does retention time correction, and all other adjustments that have been made to the xcmsSet, also the subsetting to the scanrange. That was missing in the original getEIC code for xcmsSet and caused the problem.

cheers, jo

### DIFF
--- a/R/xcmsSet.R
+++ b/R/xcmsSet.R
@@ -744,7 +744,7 @@ setMethod("group.nearest", "xcmsSet", function(object, mzVsRTbalance=10,
         mplenv$peakIdxList <- data.frame(peakidx=which(mplenv$peakmat[,"sample"]==sample),
                                          isJoinedPeak=FALSE)
         if(length(mplenv$peakIdxList$peakidx)==0){
-            cat("Warning: No peaks in sample",s,"\n")
+            cat("Warning: No peaks in sample\n")
         }
         ## scoreList <- data.frame(score=numeric(0),peak=integer(0),mpListRow=integer(0),
         ##                                                       isJoinedPeak=logical(0), isJoinedRow=logical(0))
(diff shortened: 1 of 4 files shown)


R/xcmsSet.R
```diff
@@ -744,7 +744,7 @@ setMethod("group.nearest", "xcmsSet", function(object, mzVsRTbalance=10,
         mplenv$peakIdxList <- data.frame(peakidx=which(mplenv$peakmat[,"sample"]==sample),
                                          isJoinedPeak=FALSE)
         if(length(mplenv$peakIdxList$peakidx)==0){
-            cat("Warning: No peaks in sample",s,"\n")
```
sneumann: Hi, any reason why the sample number s was dropped from the warning ?

R/xcmsSet.R
```diff
@@ -1682,15 +1682,9 @@ setMethod("getEIC", "xcmsSet", function(object, mzrange, rtrange = 200,
...
-            if(object@dataCorrection[sampidx[i]] == 1)
-                lcraw<-stitch(lcraw, AutoLockMass(lcraw))
-        }
-        if (rt == "corrected")
-            lcraw@scantime <- object@rt$corrected[[sampidx[i]]]
-        if (length(prof) > 2)
-            lcraw@profparam <- prof[seq(3, length(prof))]
+        ## getXcmsRaw takes care of rt correction, susetting to scanrage and other
+        ## stuff.
+        lcraw <- getXcmsRaw(object, sampleidx=i, rt=rt)
```
sneumann: hi @hpbenton could you please check out that the next xcms in the devel branch after merging Jo's patch still handles the weird Waters lockmass things ? If not, this should be handles properly at the getXcmsRaw -level.
Thanks, Steffen

R/xcmsSet.R
jotsetung: yes, variable `s` is not defined in the function. This would cause an error if there was no peak in the sample. Alternatively, replace `s` by `sample`, which is the sample index.

sneumann:
Merged manually and pushed to devel branch and BioC SVN.

Issue #90, opened 2016-12-07
centWave: debug me: (omz == 0)?


stanstrup:
I am trying to do some targeted peak picking manually inputting an ROI.list.

But I end up with `centWave: debug me: (omz == 0)?`
-->
https://github.com/sneumann/xcms/blob/4b2458e54e11a874fd5a0ce42cfb4d4e475a2865/R/methods-xcmsRaw.R#L420

I guess this happens when there is nothing in the mz/rt range?
Any work-around for this?

jotsetung:
Yes, looks like it stops when there is no data in that mz/rt range. 
I'm wondering what happens if instead of `stop` a simple `next` would do it too, i.e. just skipping that ROI and moving to the next one.

stanstrup:
Thanks. That seems to work. In that I get an empty table.
Does it need to find a peak in the range or should there just be data? Because it seems it errors even if the range has some spikes.

stanstrup:
Seems the recent merging of the xcms3 branch reverted this. Do I make a new pull request?

jotsetung:
Sorry my fault. You fixed the original findPeaks.centWave method that I replaced with the `do_*` function. I'll fix in the xcms3 branch, so next time Steffen merges it will be included.

jotsetung:
OK, it's now (since https://github.com/sneumann/xcms/commit/95f5c30fdd727e0419de928553121d204e032ae4) in `xcms` branch. Will be in devel once Steffen merges.

PR #91, opened 2016-12-09
changed error to warning when no peaks in ROI

stanstrup:
#90


jotsetung:
Did `R CMD check` run without problems?
I would be in favor of this change, @sneumann what do you think?

sneumann:
No objection from my side. @HendrikTreutler , does that affect your peak picking 
with predicted ROIs in any way ? Yours, Steffen

stanstrup:
@jotsetung Yeah check has no complaints (related to this).

Treutler:
I think the peak picking with predicted isotope ROIs is not affected. No objection from my side.

jotsetung:
OK, next time `xcms3` gets merged into devel it will be fixed.

PR #160, opened 2017-04-09
expand n in density to allow for effective bw

stanstrup:


jotsetung:
What I am a little puzzled is why you have this `retcor.peakgroups` method. The `retcor.peakgroups` is supposed to contain only minimal functionality as the actual code has been moved to `do_adjustRtime_peakGroups` (in the `do_adjustRtime-functions.R` file).

jotsetung:
@stanstrup , your branch is outdated. It corresponds to the code base from 16.01.2017 (check https://github.com/sneumann/xcms/network). There have been lots of things happened since then, so we should not merge your branch, otherwise we end up fixing lots of stuff.

stanstrup:
So I will attempt to process the data again and see if I can get retcor to work at all. Should I be trying with devel or xcms3?

jotsetung:
Try with `devel` please. And please send me the error message that you get with `retcor`.

stanstrup:
I tried but cannot get retcor to work. If I do the full dataset with appropiate settings I get 
```
Error in do_adjustRtime_peakGroups(peaks = peakmat, peakIndex = object@groupidx,  : 
  'minFraction' has to be between 0 and 1!
```


I tried again with just one file and default settings and I get:
```
Performing retention time correction using 752 peak groups.
Error in if (any(rtdevsmorange/rtdevrange > 2)) warn.overcorrect <- TRUE : 
  missing value where TRUE/FALSE needed
```

jotsetung:
You're using the *old* UI, right? What's your `retcor` call? What's the value for `missing` that you're using?

stanstrup:
ui?.... I guess I haven't been following closely enough changes...

My original call was:

```
xset_g = group(xset, minsamp=5, minfrac = 0.01, mzwid = 0.1, bw=2, max=300)
Processing 9196 mz slices ... OK
```

```
xset_g_r    = retcor(xset_g, plottype="deviation", span = 0.5, missing = 100, extra = 100)
Error in do_adjustRtime_peakGroups(peaks = peakmat, peakIndex = object@groupidx,  : 
  'minFraction' has to be between 0 and 1!
```


For testing with default settings I simply tried:

```
test <- xcmsSet(files=files[1])
test_g <- group(test)
Processing 3217 mz slices ... OK
```

```
test_g_r <- retcor(test_g)
Performing retention time correction using 752 peak groups.
Error in if (any(rtdevsmorange/rtdevrange > 2)) warn.overcorrect <- TRUE : 
  missing value where TRUE/FALSE needed
```

jotsetung:
Check the *new_functionality* vignette for the description of the new UI (just new function names and a slightly modified workflow).

Hm, very strange. How many samples do you have in total and per group?

stanstrup:
90 files. processed as one group.

jotsetung:
Ah, that's it. `missing` is supposed to be the number of samples that are allowed to be missing. If it was e.g. `missing = 10` all peaks would be considered that are present in `>= 80` samples (i.e. `90` total samples `- missing` samples). If you have `90` samples you can not allow peaks to be missing in `100` samples.

Funny thing is that it worked in the old code:
```r
idx <- which(nsamp >= n-missing & groupmat[,"npeaks"] <= nsamp + extra)
if (length(idx) == 0)               
    stop("No peak groups found for retention time correction")
```

with `nsamp` being the number of samples in which a peak was found and `n` being the total number of samples. In your case the first condition `nsamp >= n - missing`, i.e. `nsamp >= (90 - 100)` was always true but should have thrown an error instead.

In the new code I'm using the `minFraction` argument that is supposed to be a number between 0 and 1 specifying the proportion of samples in which a peak has to be present.

stanstrup:
IC. I think in my desperation when I couldn't separate the peaks I tried many crazy things...

I have made a new pull request with the fix in the devel branch: https://github.com/sneumann/xcms/pull/162
I added a "multiplier" of 2. I don't know if there would be a better value. But probably rare that it would make any difference.

jotsetung:
Perfect, thanks.
Are you closing this pull request then?

stanstrup:
sure.

Issue #161, opened 2017-04-09
the bandwidth of grouping is limited by n in density


stanstrup:
When the density is calculated it is using the default number of points in the returned density.
That means (I think) that the effective bw can never be more than the rtrange divided by the number of points that is by default 512. If you have a high peak capacity this can make perfectly distinct peaks inseparable. 

Here is an example. I have a 34 minutes run time GCxMS dataset. The effective bw is something like (34*60)/512 = 4 sec between datapoints I guess.

Grouping the data I cannot separate these two peaks no matter the bw:

![image](https://cloud.githubusercontent.com/assets/4532682/24837675/c22844dc-1d39-11e7-9046-66bffa8c037b.png)

There is ~5 sec between them but as you see they are not separated  (no matter the chosen bw).
I am not sure what n should be but perhaps:
`2^(ceiling(log2(diff(rtrange)/bw)))`

EDIT: perhaps `max(512, 2^(ceiling(log2(diff(rtrange)/bw))))` would be better.

That worked for me at least: 
![image](https://cloud.githubusercontent.com/assets/4532682/24837691/18c1c462-1d3a-11e7-8497-503255128dfb.png)

I made a pull request on the biocdevel branch (I couldn't get devel to play nice. Used a lot of memory and complained that minFraction was not between 0 and 1 in retcor).

https://github.com/sneumann/xcms/pull/160

Any idea if n should be some "multiple of the bw" to ensure the selected bw is possible?

jotsetung:
> I made a pull request on the biocdevel branch (I couldn't get devel to play nice. Used a lot of memory and complained that minFraction was not between 0 and 1 in retcor).

That is very strange. `biocdevel` and `devel` branches should be/are the same.

jotsetung:
For the rest, it sounds reasonable to me, especially the `max(...)`. @sneumann what do you think?

jotsetung:
Actually, I checked on some artificial data and I think @stanstrup is right. We could even increase the `n` further (e.g. by multiplying with 2), since, as far as I see it, this does not affect the overall shape of the density curve, but makes it more detailed (e.g. being able to separate neighboring peaks if `bw` is low).

The changes should however be made in the `do_chromPeaks_density` function.

In addition I would be not like to include it in the upcoming Bioc release as we would need some more testing on it.

jotsetung:
[...]

PR #162, opened 2017-04-11
fixes for #161

stanstrup:
replaces https://github.com/sneumann/xcms/pull/160

### DIFF
--- a/R/do_groupChromPeaks-functions.R
+++ b/R/do_groupChromPeaks-functions.R
@@ -114,6 +114,7 @@ do_groupChromPeaks_density <- function(peaks, sampleGroups,
 
     densFrom <- rtRange[1] - 3 * bw
     densTo <- rtRange[2] + 3 * bw
+    densN <- max(512, 2^(ceiling(log2(diff(rtRange)/(bw/2)))))
     endIdx <- 0
     num <- 0
     gcount <- integer(nSampleGroups)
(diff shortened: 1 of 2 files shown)


jotsetung:
Looks good to me. I would however not like to merge the pull request now, since we're having a *feature freeze* before the upcoming Bioconductor release (end of April).

After that I would go and merge it, but would like to have the OK from @sneumann too.

R/do_groupChromPeaks-functions.R
```diff
@@ -240,7 +241,7 @@ do_groupChromPeaks_density_par <- function(peaks, sampleGroups,
                                    sampleGroupTbl, minFr, minSmpls,
                                    sampleGroupNms, peakOrdr) {
         den <- density(z[, "rt"], bw = bw, from = rtr[1] - 3 * bw,
-                       to = rtr[2] + 3 * bw)
+                       to = rtr[2] + 3 * bw, n = max(512, 2^(ceiling(log2(diff(rtr)/(bw/2))))))
```
sneumann: Ok, n=512 is the default, and one might want to increase the number of points for large RT ranges and/or small bw ?

sneumann:
Looks good to me, and can go into the next round.

R/do_groupChromPeaks-functions.R
stanstrup: yes. If rt/n is not smaller than the requested bw (how much smaller I don't know), the bw can never be achieved because the points in the density function will be more than bw apart.
This put a de-facto limit on how small the bw could be making separating some peaks impossible to separate.

jotsetung: I think it would even not hurt increasing this number further (e.g. multiply by 2). This does not affect the shape of the density curve - just has more data points.

stanstrup: I agree. 5? I assume it couldn't need to be higher than that to get the expected result.

jotsetung:
@sneumann, are you merging the pull request? I will then merge it into xcms3

PR #164, opened 2017-04-12
Explicitly enable -O2 optimisation

### LINKED ISSUE #150: Getting rid of WARNING due to abort() down inside C++ STL
For XCMS, R CMD check is complaining in
http://bioconductor.org/checkResults/devel/bioc-LATEST/xcms/malbec2-checksrc.html

about: 
`  Found ‘abort’, possibly from ‘abort’ (C) Objects: ‘massifquant/TrMgr.o’, ‘massifquant/Tracker.o’`

This is not called in the *.cpp code, but instead from deep inside the C++ STL, which I checked running 
the compilation only up to the preprocessor, see below. I currently have no idea how avoid the `abort()` in the C++ STL. Ideas ? Yours, Steffen

```
g++ -I/usr/share/R/include -DNDEBUG      -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c massifquant/TrMgr.cpp -E | less

...
# 1770 "/usr/include/c++/5/bits/stl_list.h"
void _M_check_equal_allocators(list& __x) {
 if (_M_get_Node_allocator())
   __builtin_abort();
}
```

sneumann:
As of R-4.3 beta -O2 is not default on Linux
Without -O2, R CMD check causes a WARNING. Closes #150

### DIFF
--- a/src/Makevars
+++ b/src/Makevars
@@ -6,6 +6,9 @@ XCMSOBJECTS=fastMatch.o mzClust_hclust.o mzROI.o util.o xcms.o
 
 OBJECTS= $(MQOBJECTS) $(OBIOBJECTS) $(XCMSOBJECTS)
 
+## As of R-4.3 beta -O2 is not default on Linux.
+PKG_CXXFLAGS=-O2 
+
 all: clean $(SHLIB)
 
 clean:


jotsetung:
Awesome! Are you going to merge @sneumann ? Fine for me.

sneumann:
Does not work:
```
* checking compilation flags in Makevars ... WARNING
Non-portable flags in variable 'PKG_CXXFLAGS':
  -O2
```

Issue #184, opened 2017-06-01
Missing plots from group.density past 1.50 version (not available in 1.52)


pcm32:
Hi there!

Thanks for the excellent work!

At version 1.50 and previously, the invocation of group.density was generating plots as per this line:

https://github.com/sneumann/xcms/blob/1f816d946ed079f95dc4efa4e0aaa7cca360ef64/R/methods-xcmsSet.R#L390

This behaviour is no longer available in current versions:

https://github.com/sneumann/xcms/blob/f11aac4020319d991b845154e492f1031f946888/R/do_groupChromPeaks-functions.R#L136

While I share the spirit of removing those plot executions, I wonder whether there is anywhere else in the code base some canned call that can be used to generate these same plots from the resulting xset produced by `group`, since we want to be able to fill that output file with what was currently there (plots previously generated). Otherwise, could you offer some guidance on quick snippet that would yield those results starting with the produced xset object? It doesn't seem obvious at least that all elements required for the plot are stored in the output object.

Thanks!
Pablo

sneumann:
The plotting should be restored, to avoid breaking existing code. 
The question is where to do it. A choice could be to *not* do it in `do_groupChromPeaks_density()` 
for the sake of modularisation and keeping implicit side-effects (like plotting) to a minimum. 
In that case, it could be restored (or called from) https://github.com/sneumann/xcms/blob/f11aac4020319d991b845154e492f1031f946888/R/methods-xcmsSet.R#L343 
where we still have the `sleep` parameter that is controlling the plotting.

jotsetung:
Was my fault. I removed the plotting functionality from the analysis algorithms because I wanted to separate the two things. I'll look into adding the visualizations again for the identified groups.

pcm32:
Do you plan to make a new release to BioConductor when this is reinstated? To know roughly from where and when we could access the functionality. Thanks!

jotsetung:
We will push the changes definitely to the Bioconductor devel branch - I'm not sure if we're going to add them also to the current Bioconductor release branch (i.e. Bioconductor 3.5).
Regarding timing I can't promise anything as I'll be on holidays soon - but should be done by Mid July - or earlier. I'll let you definitely know when it's done.

jotsetung:
@pcm32 , the `sleep` parameter is working again in the `xcms` version 2.99.2 available in the current Bioconductor development version.
Alternatively, you can install it also from github:
```r
library(devtools)
install_github("sneumann/xcms")
[...]

PR #185, opened 2017-06-05
Fix issue #184

jotsetung:
- Add parameter sleep to the do_groupChromPeaks_density function to ensure
  backward compatibility with the old group.density code.


sneumann:
Hi, patch looks good to go, graphics are restored and I checked on faahKO here.

Issue #235, opened 2017-11-07
Adding weighted moving average to spectrum smoothing


SiggiSmara:
A pull request is already submitted to [MALDIquant](//github.com/sgibb/MALDIquant) to add weighted average smoothing. I'm also adding a pull request to here and to MSnbase so that people can take advantage of it once it is merged.

jotsetung:
Closing this issue since it has been addressed in `MSnbase`/`MALDIquant`.

PR #236, opened 2017-11-07
Adding support for weighted moving average from MALDIquant

SiggiSmara:
As explained in #235


sgibb:
Depends on https://github.com/sgibb/MALDIquant/pull/54 and https://github.com/lgatto/MSnbase/pull/273/.

SiggiSmara:
Looking at the code and the changes suggested in sgibb/MALDIquant#54 I think my changes here are actually not needed any more since the weighted moving average is being folded into the moving average function.

jotsetung:
In case that's not needed anymore feel free to close @SiggiSmara

SiggiSmara:
Not needed any more

PR #282, opened 2018-05-15
peans_ncols -> peaks_ncols typo

stanstrup:
I ran into an error with this and wondered what peans was ;)


jotsetung:
YOU DON'T KNOW WHAT peans IS??? ;)
thanks a lot! that slipped through - and i was apparently lazy in checking all arguments in the unit tests 😊

PR #376, opened 2019-05-08
Michael devel

michaelwitting:
functions for the alignment and correlation

### DIFF
--- a/R/functions-xcmsSwath.R
+++ b/R/functions-xcmsSwath.R
@@ -1,11 +1,14 @@
 ## function "align" EICs
-.alignEics <- function(x, y) {
+# are the MS1 chroms, y are the MS2
+.align_chromatogram <- function(x, y, na.value = NA) {
 
   # linear interpolation of y on x
   y_aligned <- approx(rtime(y), intensity(y), rtime(x))
   
   # deal with NA
-  y_aligned[is.na()] <- 0
+  if(!is.na(na.value)) {
+    y_aligned[is.na()] <- na.value
+  }
   
   # correct rtime and int in chromatogram object
   y@rtime <- y_aligned$x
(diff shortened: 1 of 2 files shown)


R/functions-xcmsSwath.R
```diff
@@ -0,0 +1,33 @@
+## function "align" EICs
+.alignEics <- function(x, y) {
+
+  # linear interpolation of y on x
+  y_aligned <- approx(rtime(y), intensity(y), rtime(x))
+  
+  # deal with NA
+  y_aligned[is.na()] <- 0
```
jorainer: I would rather keep the `NA`s . Otherwise a subsequent correlation would correlate an intensity with a 0, if it was `NA` (because there is no signal for that specific retention time) the value pair could be skipped.

R/functions-xcmsSwath.R
```diff
@@ -0,0 +1,33 @@
+## function "align" EICs
+.alignEics <- function(x, y) {
```
jorainer: Sorry for being picky, but could you rename that to `.align_chromatogram` or `.alignChromatogram` instead? Then the user knows that he/she is aligning chromatograms.

R/functions-xcmsSwath.R
```diff
@@ -0,0 +1,33 @@
...
+  y@rtime <- y_aligned$x
+  y@intensity <- y_aligned$y
+  
+  # return
+  return(y)
+  
+}
+
+# function to correlate EICs
+.correlateEics <- function(x, y) {
```
jorainer: Also here, IMHO it would be better to call that `.correlate_chromatogram` or `.correlateChromatogram`.

R/functions-xcmsSwath.R
```diff
@@ -0,0 +1,33 @@
...
+  
+  # check if x and y are on the same time scale
+  if(!rtime(x) == rtime(y)) {
+    y <- .alignEics(x, y)
+  }
+  
+  # correlate values
+  corValues <- cor.test(intensity(x), intensity(y))
```
jorainer: Do we use also the p-values? I think we can live with *just* the correlation coefficient and then we could use `cor` instead and return a single value instead of a `list` (would make later processing easier of the result also).

R/functions-xcmsSwath.R
jorainer: I would also add here parameters of the `cor` function, i.e. `use = "pairwise.complete.obs"` and `method = c("pearson", "kendall", "spearman")`. These should then be passed along to the `cor` function.

jorainer:
And one more thing: please add also unit tests for these functions. One for each function and put them into *tests/testthat/test_functions-xcmsSwath.R*

R/functions-xcmsSwath.R
michaelwitting: True. So far I always replaced them with zeros. Can we agree to have an option like `na.replace` and if this is `TRUE` than we have `0` otherwise `NA`. Default value will be `FALSE`

R/functions-xcmsSwath.R
michaelwitting: No problem.

michaelwitting: Changed to `.align_chromatogram`

R/functions-xcmsSwath.R
michaelwitting: No problem.

michaelwitting: You want a `...` notation or explicit?

R/functions-xcmsSwath.R
michaelwitting: I used them sometimes. Agree, single value is easier to handle

R/functions-xcmsSwath.R
jorainer: Sure. And what about an option `naValue = NA` and replace `NA` with that value (only `if (!is.na(naValue)) y_aligned[is.na(y_aligned)] <- naValue`)?

michaelwitting:
Wil make the changes during the day and commit and push later. I would love to add the unit tests, but my R installation is not completely working at the moment. Will do that once I'm back to normal (whatever that means).

R/functions-xcmsSwath.R
jorainer: 👍

michaelwitting:
What file for unit tests? The SWATH file from the Metabolomics 2018 presentation?

R/functions-xcmsSwath.R
jorainer: I would add them as explicit parameters.

jorainer:
For these functions I'd suggest you create some bogus `Chromatogram` objects. So you have also more control over their values.

michaelwitting:
Tried my best ;-)

michaelwitting:
There is also a place holder for a function to deal with `Chromatograms` instead of single `Chromatogram`. I'm not familiar with the `biocParallel` backend, so you might have a good idea for a fast implementation

jorainer:
Excellent! Thanks! I will just slightly modify some stuff (line break at 80 characters and similar) and eventually extend the unit tests.

PR #385, opened 2019-06-05
Update xcms.Rmd

Maddocent:
Fix to the do.call part at line 210, more robust code. This part of the code failed for my Example using my own data (LC/MS/MS). This fix solved the problem. 

Hope it helps,

Kind regards,
Marc Teunis

### DIFF
--- a/vignettes/xcms.Rmd
+++ b/vignettes/xcms.Rmd
@@ -207,7 +207,7 @@ binned base peak chromatograms.
 bpis_bin <- bin(bpis, binSize = 2)
 
 ## Calculate correlation on the log2 transformed base peak intensities
-cormat <- cor(log2(do.call(cbind, lapply(bpis_bin, intensity))))
+cormat <- cor(dplyr::bind_cols(lapply(bpis_bin, intensity)))
 colnames(cormat) <- rownames(cormat) <- raw_data$sample_name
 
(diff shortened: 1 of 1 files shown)


jorainer:
Thanks for the PR @Maddocent !

What was the error that you got? Would be nice if this could be fixed without requiring the `dplyr` package, as `xcms` has already too many dependencies IMHO.

vignettes/xcms.Rmd
```diff
@@ -207,7 +207,7 @@ binned base peak chromatograms.
 bpis_bin <- bin(bpis, binSize = 2)
 
 ## Calculate correlation on the log2 transformed base peak intensities
-cormat <- cor(log2(do.call(cbind, lapply(bpis_bin, intensity))))
+cormat <- cor(dplyr::bind_cols(lapply(bpis_bin, intensity)))
```
sneumann: Hi, saw the PR only now. So you intend to replace `do.call(cbind)`
with `bind_cols()`, but your patch is dropping the log2. Yours, Steffen

jorainer: Also, I would prefer a patch that does not add new dependencies - and some information or example data why/where the original code failed

sneumann:
Hi, without a reproducible issue I'd like to close this PR. Yours, Steffen

Issue #397, opened 2019-07-27
Typo in Tutorial


SengaSacul:
Hi there,

might it be a typo in the  Tutorial ?

Function adjustedRtimePeakGroups is not existing, guess should be adjustRtimePeakGroups, to get the matrix ;)

PR #398, opened 2019-07-27
Fix typo

sneumann:
Thanks to @SengaSacul for reporting in #397


jorainer:
Thanks @sneumann and @SengaSacul

Issue #410, opened 2019-09-23
Strange behavior with findChromPeaks for a single Chromatogram object


wkumler:
Hi all,
I've been using xcms for a while now and really loving it, but have only recently started to look into the methods behind the magic. 

One thing I'm noticing is that xcms struggles to identify peaks if there's only a single local maximum present in the entire spectrum. For example:

```r
library(xcms)

scans <- 201:400
# Generate a nice Gaussian distribution for our peak
peak_template <- as.numeric(table(cut(rnorm(10000), breaks = 20)))
# Buffer the peak on both sides with zeros (could be any small number, though)
single_peak_ints <- c(rep(0, 90), peak_template, rep(0, 90))

chrom <- Chromatogram(rtime = scans, intensity = single_peak_ints, mz = 250)

findChromPeaks(chrom, param = CentWaveParam())
```

returns the following output:

> Object of class: XChromatogram
> length of object: 200
> from file: 
> mz range: [250, 250]
> rt range: [201, 400]
> MS level: 1
> Identified chromatographic peaks (0):
>  rt	rtmin	rtmax	into	maxo	sn (1 more column(s))
> Warning message:
> In peaksWithCentWave(int = c(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,  :
>   No peaks found!

implying that there are no peaks in the model spectrum.

Some additional digging turned up the reason for this. `findChromPeaks` calls `peaksWithCentWave` calls `.getRtROI`, which in turn uses MALDIquant's `.localMaxima`  to identify local maxima. These local maxima are then used to determine the scan numbers of the regions of interest, which are passed to the prefilter check.

However, if there's only a single local maxima identified, then the prefilter check changes its behavior and iterates over each data point individually, rather than the peak as a whole:

```r
# Use the single peak spectrum from above
int <- single_peak_ints
# Manually add in parameters from CentWaveParam()
peakwidth <- c(20, 50)
prefilter <- c(3, 100)
noise <- 0
# Manually add parameters produced by call to .getRtROI()
rt_step <- 1 # Assume each scan takes exactly one second
up_bound <- ceiling(peakwidth[2]/rt_step)

# Find local maxima
pk_idx <- which(MALDIquant:::.localMaxima(int, floor(peakwidth[1]/rt_step)))
scmin <- sapply(pk_idx - up_bound, max, y = 1)
scmax <- sapply(pk_idx + up_bound, min, y = length(int))
[...]

PR #411, opened 2019-09-24
Resolved issue #410 by adding SIMPLIFY = F argument to findChromPeaks

wkumler:
Per @jorainer's recommendation, this pull request resolves issue #410  where peaks were unable to be found because `mapply` was simplifying to a matrix, which changed the subsequent prefilter check to reject ROIs if they were all the same length.

### DIFF
--- a/R/do_findChromPeaks-functions.R
+++ b/R/do_findChromPeaks-functions.R
@@ -3592,7 +3592,7 @@ peaksWithCentWave <- function(int, rt,
     scmin <- sapply(pk_idx - up_bound, max, y = 1)
     scmax <- sapply(pk_idx + up_bound, min, y = int_len)
     ## Second filter: at least k values larger I
-    roi_idxs <- mapply(scmin, scmax, FUN = seq)
+    roi_idxs <- mapply(scmin, scmax, FUN = seq, SIMPLIFY = F)
     ok <- vapply(roi_idxs,
                  FUN = function(x, k, I) {
                      sum(int[x] >= I) >= k
--- a/R/do_findChromPeaks-functions.R
+++ b/R/do_findChromPeaks-functions.R
@@ -3592,7 +3592,7 @@ peaksWithCentWave <- function(int, rt,
     scmin <- sapply(pk_idx - up_bound, max, y = 1)
     scmax <- sapply(pk_idx + up_bound, min, y = int_len)
     ## Second filter: at least k values larger I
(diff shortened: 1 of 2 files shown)


sneumann:
Hi WIlliam, could you augment your PR with a unit test case ?
E.g. something like the example you have in #410 ?
Maybe one or two cases more while at it ? One where ROIs 
should be detected, and some which just do (and do not) pass the `function(x, k, I)` test ?
That way we'll also extend the test coverage ;-) Yours, Steffen

wkumler:
@sneumann Done! I've added four test cases - one with a single peak that's successfully found, one with three peaks each with their own ROI, and two with varying prefilter settings that select for one or two peaks out of the three. I used the same modeled data from the #410 examples because I wasn't able to track down the faahKO file that the other test cases use. That also means that I wasn't able to check the previous test cases for compatibility with the new settings, sorry. Hopefully those test cases don't rely on not finding peaks that will now be found!

R/do_findChromPeaks-functions.R
```diff
@@ -3592,7 +3592,7 @@ peaksWithCentWave <- function(int, rt,
     scmin <- sapply(pk_idx - up_bound, max, y = 1)
     scmax <- sapply(pk_idx + up_bound, min, y = int_len)
     ## Second filter: at least k values larger I
-    roi_idxs <- mapply(scmin, scmax, FUN = seq)
+    roi_idxs <- mapply(scmin, scmax, FUN = seq, SIMPLIFY = F)
```
jorainer: Please use `SIMPLIFY = FALSE` instead of `F`.

jorainer (changes requested):
Thanks @wkumler for the great PR (also adding the unit test)! 
There is only one thing I'd like to have changed.

R/do_findChromPeaks-functions.R
wkumler: Done! Good call.

jorainer:
I'll Merging now - I believe this is a really great fix for peak detection in chromatographic data. Thanks again @wkumler !

sneumann:
I can confirm that it passes `R CMD check` also here. 
In case you want to use the faahKO files in the future, see https://github.com/sneumann/xcms/blob/24471b789ff4486688f0ba2aa1ac3373d93f38b7/tests/testthat.R#L18
Thanks for the contribution, yours, Steffen

wkumler:
Cheers! Glad it worked out.

PR #423, opened 2019-10-26
Update methods-XCMSnExp.R

mengchen18:
when ftIdx[[i]] (peakidx) is length 1, the resulted cur_pks would be a vector, not a matrix as expect. The next line "cur_pks[, value]" would throw an error message.


jorainer:
Thanks @mengchen18 for the PR!

PR #446, opened 2020-01-21
Added Bioschemas annotation to the xcms.Rmd vignette

egonw:
Adds Bioschemas annotation for one of the vignettes. I suggest one *.bioschemas.html for each vignette. Note how toe xcms.Rmd YAML points to the `xcms.bioschemas.html` file, which contains only a `<script>` element with JSON-LD in it.

### DIFF
--- a/vignettes/xcms.Rmd
+++ b/vignettes/xcms.Rmd
@@ -4,6 +4,8 @@ package: xcms
 output:
   BiocStyle::html_document:
     toc_float: true
+    includes:
+      in_header: xcms.bioschemas.html
 vignette: >
   %\VignetteIndexEntry{LCMS data preprocessing and analysis with xcms}
   %\VignetteEngine{knitr::rmarkdown}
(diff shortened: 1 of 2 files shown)


jorainer:
How does that then work? Is this html file then included into the generated vignette's html file? And you are then parsing the htmls of all Bioconductor packages for these entrys?

egonw:
@jorainer, yes, this HTML snippet is included in the file. When you do a `R CMD build` and extract the xcms.html, search for the string `ld+json` and you will find the snippet included. The two extra lines in `xcms.Rmd` tell the build system to include it.

Yes, TeSS uses crawlers that will look for `ld+json` snippets. And so will Google and Bing, actually.

jorainer:
thanks for the clarification @egonw

egonw:
> Would it be possible to call that additional file "xcms.bioschemes.yml" (or some other file ending) instead of *.html?

I think so, but it is not a YAML file.

jorainer:
It's fine. You're right, html makes most sense. I've merged.

Issue #445, opened 2020-01-20
Reducing "minimum peakwidth" parameter removes peaks during wavelet transform


wkumler:
XCMS seems to be highly sensitive to the minimum peak width when finding peaks without a lot of data outside of the peak itself. If the minimum peakwidth is set too low, the peaks themselves are dropped without warning.

It looks like this happens because the `scales` argument used by `.centWave_orig` function is determined by the minimum peakwidth, but the scales themselves are only applied to a chromatogram of a certain length.

From `xcms:::centwave_orig()`:

``` r
peakwidth <- c(20, 50) #or whatever the user set the CentWaveParam() to

# A few lines later...
scalerange <- round((peakwidth)/2)

# A few lines beyond that...
# Assuming that a min and max were provided, as seems to be the typical usage
scales <- seq(from = scalerange[1], to = scalerange[2], by = 2) 
```

Now, `scales` is passed directly to the wavelet function `MSW.cwt()`:

`wCoefs <- MSW.cwt(d, scales = scales, wavelet = "mexh")`

which uses it to determine the widths of the wavelets to test against the peak.

However, if there aren’t enough scans in the extracted chromatogram (`d`, which becomes `ms` within the `MSW.cwt()` function), then the wavelets aren’t run on the chromatogram and instead `NA` is returned.

``` r
# If there's a narrow buffer on each side (less than 10 scans)
actual_peakwidth <- 35
perfect_peak <- dnorm(seq(-3, 3, length.out = actual_peakwidth))
buffer <- numeric(10)
sample_chrom <- c(buffer, perfect_peak*100, buffer) + runif(length(buffer)*2+actual_peakwidth)
plot(sample_chrom, type="b", ylab="Peak intensity", xlab="Retention time (s)",
     main="Chromatogram with very few scans on each side")
#User-provided peakwidth (this is the default)
peakwidth <- c(20, 50)
#Say the scans are 1 second apart
scalerange <- round((peakwidth/1)/2)
scales <- seq(from = scalerange[1], to = scalerange[2], by = 2)
wCoefs <- xcms:::MSW.cwt(sample_chrom, scales = scales, wavelet = "mexh")
head(wCoefs)
# Returns NA, meaning the peak was not found

# If the buffer is increased slightly (to 15 scans)
buffer <- numeric(15)
sample_chrom <- c(buffer, perfect_peak*100, buffer) + runif(length(buffer)*2+actual_peakwidth)
[...]

PR #449, opened 2020-02-10
Resolved issue #445 by changing MSW.cwt's extend length method

wkumler:
Per @jorainer's recommendation, this pull request resolves issue #445 where peaks were dropped from the peaklist during wavelet transformation because the minimum peakwidth was too low. The new behavior also ensures that all wavelet scales passed to the MSW.cwt function are actually run, rather than being selected via weird interaction between the minimum peakwidth and the number of scans in the to-be-transformed chromatogram.


R/cwTools.R
```diff
@@ -22,7 +22,13 @@ MSW.cwt <- function (ms, scales = 1, wavelet = "mexh")
         stop("Unsupported wavelet!")
     }
     oldLen <- length(ms)
-    ms <- MSW.extendNBase(ms, nLevel = NULL, base = 2)
```
jorainer: instead of removing the old with your new code I would suggest to have here something like a:
```r
if (extendLengthMSW) {
    your new code
} else
old code
```

wkumler:
Hi @jorainer, sorry for the delay! I've added two commits that enable extendLengthMSW as a parameter within the CentWaveParam that is carried through, preserving the original peakfinding if set to FALSE (the default). I had to append a couple additional files to get it to compile correctly, so it's probably worth taking an extra look at the second commit to make sure it's all still up to scratch.

wkumler:
@jorainer Done and done! I ended up manually editing the DESCRIPTION because I couldn't find the roxygen source for it - is that auto-generated?

wkumler:
No worries! This is my first time contributing to a large package like this, so your willingness to work with me means a lot. I believe I've updated my fork to the current master, added some documentation and ran `devtools::document()` to generate the `.Rd` files, and updated the unit tests to reflect the changed code. I still got some errors when running `devtools::check()`, mostly relating to BiocParallel not playing nicely with 32-bit R on Windows (namespace functions not found because they aren't passed to the SNOW workers explicitly). But `devtools::test()` now runs without any errors and it compiles completely on my end.

PR #469, opened 2020-04-27
tests: add longtests

jorainer:
- Move unit tests for the *old* xcms classes to the longtests folder.


.Rbuildignore
```diff
@@ -8,3 +8,4 @@ local_data
 README.md
 Rplots.pdf
 .DS_Store
+.BBSoptions
```
sneumann, line 11: Why add to .gitignore ? These are control files for BioC build system, so it'd need to arrive there,

tests/testthat/test_do_adjustRtime-functions.R
```diff
@@ -1,9 +1,9 @@
 test_that("getPeakGroupsRtMatrix works", {
     param <- PeakGroupsParam()
     nSamples <- length(fileNames(xod_xg))
-    pkGrp <- .getPeakGroupsRtMatrix(
+    pkGrp <- xcms:::.getPeakGroupsRtMatrix(
```
sneumann, line 4: So all non-exported functions in tests need to be prefixed ?

.Rbuildignore
jorainer: I was not sure - `R CMD check` complains about this file. But you're right, I will remove that line again.

tests/testthat/test_do_adjustRtime-functions.R
jorainer: No - that's a leftover when I run the tests line by line locally.

.Rbuildignore
jorainer: I removed this in the latest commit.

jorainer:
> Can we have instructions on longtests / linkt to the BioC page somewhere

what exactly do you mean? Once the first *long test* was performed we could link that page in the README.md.

sneumann:
lgtm. Yours, Steffen

PR #494, opened 2020-06-24
Adapt to MSnbase 2.15.3

jorainer:
Adapt to the recent changes in `MSnbase` that involved renaming of `Spectra` to `MSpectra` and `Chromatograms` into `MChromatograms`.


jorainer:
A well, since MSnbase 1.15.3 is not yet available unit tests fail. Will merge anyway.

PR #532, opened 2021-01-02
Import ProtGenerics::quantify (was MSnbase::quantify)

lgatto:
Note: GitHub action is likely to fail because ProtGenerics 1.23.6 is available on GitHub and was just pushed to Bioc.


jorainer (approved):
Thanks @lgatto !

PR #543, opened 2021-02-22
Export findmzROI method

jorainer:


sneumann:
Hm, the function has no docs yet `Undocumented code objects: ‘findmzROI’`
which admittedly is easy to miss amidst all the other output. But that'd get us a `WARNING` on BioC.
Yours, Steffen

jorainer:
yep - just added the documentation.

PR #555, opened 2021-03-30
Feature/plot q cxcms2

sneumann:
Since only three function calls at the beginning differ between xcmsSet and XCMSnExp,
I kept plotQC() as a plain function with class(object) to distinguish between the cases.
Yours,
Steffen


jorainer (approved):
Looks good to me. Thanks @sneumann

Issue #564, opened 2021-06-14
Build error and tons of "No data points ... Returning empty spectrum."


sneumann:
```
Loading required package: GenomeInfoDb
Polarity can not be extracted from netCDF files, please set manually the polarity with the 'polarity' method.
Warning in trimMz_Spectrum(object, mzlim = mz, msLevel. = msLevel., ...) :
  No data points between 334.9 and 335.1 for spectrum with acquisition
number 128. Returning empty spectrum.
Warning in trimMz_Spectrum(object, mzlim = mz, msLevel. = msLevel., ...) :
  No data points between 334.9 and 335.1 for spectrum with acquisition
number 131. Returning empty spectrum.
Warning in trimMz_Spectrum(object, mzlim = mz, msLevel. = msLevel., ...) :
  No data points between 334.9 and 335.1 for spectrum with acquisition
number 132. Returning empty spectrum.
```

jorainer:
The warnings come from the `filterMz` (`trimMz`) function from `MSnbase`. Everything above is *only* a warning, so that should not cause a build error.

jorainer:
Seems that the error comes from the `estimatePrecursorIntensity` which is now also available in the `Spectra` package. I could fix this quickly if OK.

PR #565, opened 2021-06-16
Fix build errors (issue #564)

jorainer:


jorainer:
Regarding 4.1 - this `xcms` is the current devel version for BioC 3.14 - and that should run on 4.1. I think we should try to avoid as much as possible running newer `xcms` versions on old R and BioC packages.

sneumann:
Hm, not convinced. We should of course aim for the correct versions, and indeed with conda and containers it has become easier for users to stay on current versions, but unless technically necessary/beneficial I wouldn't force users to install current versions, not even supporting N-1. And, by allowing N and N-1 we might even be more robust. And of course, if necessary we bump dependency versions, but I wouldn't count the vignette there. Yours, Steffen

jorainer:
I changed it back to `magrittr` - is it then OK for you if I merge it @sneumann ?

PR #572, opened 2021-07-20
Add LC-MS feature grouping functionality to xcms

jorainer:
- Add LC-MS feature grouping functionality based on the `MsFeatures` package to `xcms`.
- Add a vignette describing this grouping.
- Adapt to recent changes in `ProtGenerics` and `MSnbase` (some methods were moved from `MSnbase` to `ProtGenerics`).


jorainer:
Note that the unit tests fail on macOS and Windows because of the warning of `mzR` being built with a different `Rcpp` version.

R/methods-MChromatograms.R
```diff
@@ -354,3 +277,309 @@ setMethod("filterColumnsKeepTop", "MChromatograms",
...
+#' ## plot function for chromatographic data, which will draw the EICs for
+#' ## multiple samples into the same plot
+#' plot(chrs)
+#'
+#' ## Converting the object to a MChromatograms without detected peaks
+#' chrs <- as(chrs, "MChromatograms")
+#'
+#' plotChromatogramsOverlay(chrs,
+#'     col = c("#ff000040", "#00ff0040", "#0000ff40"))
+setMethod("plotChromatogramsOverlay", "MChromatograms",
```
sneumann, line 416: I was wondering if optional log-scaling and/or normalisation would be interesting, to better reveal small peak shapes (e.g. 2nd or 3rd isotope ...)

R/methods-XCMSnExp.R
```diff
@@ -361,7 +361,7 @@ setMethod("chromPeaks", "XCMSnExp", function(object, bySample = FALSE,
                    drop = FALSE]
     ## Select peaks within rt range.
     if (length(rt)) {
-        rt <- range(rt)
+        rt <- range(as.numeric(rt))
```
sneumann, line 364: Why was that conversion needed ? What "broken" rt did you observe ?

R/methods-XCMSnExp.R
```diff
@@ -1129,14 +1129,14 @@ setMethod("[[", "XCMSnExp",
 #'     parent class.
 #'
 #' @author Johannes Rainer
-setMethod("bin", "XCMSnExp", function(object, binSize = 1L, msLevel.) {
-    if (hasAdjustedRtime(object) | hasFeatures(object) |
-        hasChromPeaks(object)) {
+setMethod("bin", "XCMSnExp", function(x, binSize = 1L, msLevel.) {
```
sneumann, line 1132: was there a need to rename object to x ?

R/methods-group-features.R
```diff
@@ -0,0 +1,1049 @@
+#' @title Grouping of LC-MS features
```
sneumann, line 1: `Grouping of LC-MS features` is quite generic and overloaded term. /me is wondering if we should dare what we didn't in CAMERA and call 'em compound spectra containing groups of peaks ....

R/methods-group-features.R
```diff
@@ -0,0 +1,1049 @@
...
+        else is_msLevel <- rep(TRUE, nrow(featureDefinitions(object)))
+        if (all(is.na(fgs)))
+            fgs <- rep("FG", length(fgs))
+        fgs[!is_msLevel] <- NA
+        nas <- is.na(fgs)
+        fgs <- factor(fgs, levels = unique(fgs))
+        rtl <- split(featureDefinitions(object)$rtmed, fgs)
+        res <- lapply(
+            rtl, function(z, param)
+                MsFeatures:::.format_id(groupFeatures(z, param = param, ...)),
```
sneumann, line 176: Should that eventually be exported by MsFeatures ?

R/methods-group-features.R
```diff
@@ -0,0 +1,1049 @@
...
+        validObject(object)
+        object
+    })
+
+#' @title Group EICs based on their correlation
+#'
+#' @description
+#'
+#' `groupEicCorrelation` groups (extracted ion) chromatograms (EICs) based on
+#' their similarity with each other. If this correlation is `>=` than the
```
sneumann, line 973: I sometimes contemplate whether we need both the correlation threshold and the p-Value to avoid false positive (high) correlations based on too few scans...

sneumann (commented):
Hi, I think my main point is the naming, since the term "grouping" is quite overloaded, a mistake we might've started back in CAMERA ... 
Are we now "compounding" features ? Likely not "Combining" nor "aggregating" ? 
While superficially only wording on the brink of bike-shedding, it could confuse new users. 
Yours, Steffen

R/functions-XChromatogram.R
```diff
@@ -203,15 +203,17 @@ XChromatogram <- function(rtime = numeric(), intensity = numeric(),
...
-               points(pks[, "rt"], pks[, "maxo"], pch = pch, col = col,
-                      bg = bg, ...)
+               points(pks[, "rt"], pks[, "maxo"] + yoffset, pch = pch,
+                      col = col, bg = bg, ...)
            },
            rectangle = {
                rect(xleft = pks[, "rtmin"], xright = pks[, "rtmax"],
-                    ybottom = rep(0, nrow(pks)), ytop = pks[, "maxo"],
+                    ybottom = rep(yoffset, nrow(pks)),
+                    ytop = pks[, "maxo"] + yoffset,
```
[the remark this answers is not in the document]
jorainer: this functions highlights identified chromatographic peaks in an EIC. The `"maxo"` is hardcoded because in most cases either the apex position (`type = "point"`) or a rectangle or polygon is drawn that has this apex position as its *highest* point. 
If you suggest to add an additional parameter that would allow the user to choose this, that would be tricky IMHO, because it would only work for `type = "point"` and `type = "rectangle"` but not for the other type(s).

R/functions-xcmsSwath.R
```diff
@@ -172,8 +172,8 @@
         return(.data2spectra(fromFile = fromFile, rt = x["rt"], polarity = pol,
                              precursorIntensity = x[column],
                              precursorMz = x["mz"], return.type = return.type))
-    cors <- vapply(chr_2@.Data, .correlate_chromatogram, y = chr_1[1, 1],
-                   numeric(1), align = "approx")
+    cors <- vapply(chr_2@.Data, compareChromatograms, y = chr_1[1, 1],
+                   numeric(1), ALIGNFUNARGS = list(method = "approx"))
```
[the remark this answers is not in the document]
jorainer: Good point. R and BioC don't really agree on that and there is no complete consistency. I also don't really like the ALLCAPS but it's e.g. also used in `lapply` (`X`, `FUN`), `apply` (`MARGIN`). We started using the ALLCAPS thus also in `Spectra` e.g. in `compareSpectra` where we have `FUN` and `MAPFUN` that allow to define functions to be used for similarity calculation respectively mapping of peaks (a la `lapply`).

Now, the problem with `compareSpectra` is that we have `...` to pass additional arguments to **both** functions, the `FUN` and `MAPFUN`. It kinda works, but can be tricky if a parameter with the same name, but different value should be passed to each of the two functions. Thus, the `compareChromatograms` function has parameters `ALIGNFUN` for aligning of the chromatograms and `FUN` to calculate similarity but in addition also the parameters `ALIGNFUNARGS` and `FUNARGS` that allow the user to define arguments specifically for each of the two functions. It's not pretty but efficient...

My (kind of) *consistency* in this inconsistency: 
- I use camelCase argument namings if the argument is used within that function (the default)
- ALLCAPS is used e.g. if a function is expected as the argument to be applied to the data (a la `lapply`)  
- `PARAM` *vs* `ARGS`: for me `PARAM` (or `Param`) implicates a single parameter, and I used it almost exclusively if a parameter **object** is expected (such as `BPPARAM` or `CentWaveParam`). Secondly, I would prefer `ARGS` over `PARMS` because it IMHO reads better - and might be easier to understand for the user.

R/methods-Chromatogram.R
```diff
@@ -149,6 +149,8 @@ setMethod("findChromPeaks", signature(object = "Chromatogram",
 #'
 #' @description
 #'
+#' **Please use [compareChromatograms()] instead of `correlate`**
```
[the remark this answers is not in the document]
jorainer: Thanks, I will add that.

R/methods-MChromatograms.R
jorainer: Good point. I will add a parameter `transform = identify` that would allow the user to e.g. specify `transform = log` to log transform the data.

R/methods-XCMSnExp.R
jorainer: this is just a failsafe - we had cases in which a `character` was passed with `rt` - and the function did neither throw an error nor show a warning.

R/methods-XCMSnExp.R
jorainer: yes, because `bin` is now imported from `ProtGenerics` - there were multiple `bin` definition and now this was unified into one - with `x` instead of `object`.

R/methods-group-features.R
jorainer: yes/no. For now it's a private function there, but it's a valid point. I will consider/rethink that.

R/methods-group-features.R
jorainer: totally agree. But what I did **not** want to happen here is that we restrict ourselfs to just Pearson's correlation for similarity calculation (i.e. using `cor.test` instead of `cor` which then calculates also the p-value). I wanted to keep the similarity calculation as customizable as possible: we can use any function that returns a numeric value as similarity function. Instead of a p-value I would then maybe require a minimum number of matched/valid data points?

R/methods-group-features.R
jorainer: Agree, what if we use *Compounding of LC-MS features* as title here?

R/functions-XChromatogram.R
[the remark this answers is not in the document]
sneumann: indeed, so leave as-is.

R/functions-xcmsSwath.R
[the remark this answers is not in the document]
sneumann: Ok

R/methods-group-features.R
sneumann: Sounds good to me. Maybe we can run the idea past a few other contributors as well for confirmation.

R/methods-group-features.R
sneumann: I agree to keep things simple (and generic). So a user that wants a second threshhold (i.e. <= p) needs to implement `correlationP0995()` overwriting any standard values. Perfectly fine.

jorainer:
Thanks @sneumann for your comments! I made now some changes (e.g. add parameter `transform` to `plotChromatogramsOverlay` to allow data transformation).

Regarding the naming: I totally agree. I would keep the names of the functions as they are (i.e. `groupFeatures`) because they are defined in `MsFeatures` and actually are just defining groups - anything like `combine` `aggregate` would imply IMHO already a combination of the actual values. If OK for you I would do some rewording in the documentation and the vignette to replace *feature grouping* with e.g. the *compounding* that you suggested (is that in fact a recognized word in untargeted LC-MS analysis?).

R/methods-group-features.R
jorainer: I would not like to rename the `groupFeatures` method as this method (defined in `msFeatures`) is thought to be as generic as possible - it's just a concept of allowing to *group MS features* - in our case here we're trying to group features because they might be from the same compound, but there could also be other use cases - also thinking of proteomics (grouping of peptides to proteins). Thus, I would like to keep the name of the method as it is, but clearly describe the goal of this grouping in `xcms` (i.e. *compounding*).

jorainer:
@sneumann , I've reformulated part of the documentation and vignette also introducing the term *compounding*. Please have a look at it if you find the time.

sneumann:
looks good to me now.

PR #581, opened 2021-10-08
Small updates and unit test fix

jorainer:


jorainer:
@sneumann , this PR should fix the build errors. The errors on macOS and Windows are coming from the warning that mzR was build with a different version of Rcpp

Issue #603, opened 2022-03-13
Error: No MS level 2 spectra present


13479776:
Hi, all,

   I use the xcms3 to get the XCMSnExp object with dataset including full scan or DDA mode.  The table(XCMSnExp) function can give me the MS1 and MS2 scan information.

   However, when i use the  chromPeakSpectra() function, the error "Error: No MS level 2 spectra present" was happened.
   Is it possible to process and got the (MS2) spectra  associated with the features the dataset with mixture scan mode? such as mzML with full scan and mzML with DDA mode.

bes wishes,
hees

PR #604, opened 2022-03-14
Jomain

jorainer:
This PR adds:
- `transformIntensity` function.
- fix issue #603


jorainer:
@sneumann , can you please have a look at this PR, does not add much new things but fixes a bug/issue. If OK, please merge and push to BioC (along with the RELEASE_3_14 branch that contains the same fix).

sneumann:
Looks good, and passes checks here. There is no unit test for #603, could one assemble such from the existing MS files ? Yours, Steffen

jorainer:
you're right, there is none. AFAIK we don't have any (meaninful) MS1 and MS1+MS2 data in `msdata` we could use for that. I did test this locally on some of my data, but that's obviously not a *real* unit test.

sneumann:
Maybe we can assemble a less meaningful test from existing mzMLs in msdata ?

Issue #590, opened 2021-12-16
when more than one m/z is inside the ROI the m/z can be way off


stanstrup:
I encountered an issue that happens when more than one m/z value is inside the ROI for a scan. It seems that what XCMS does is take the average m/z value and sum the intensities of all peaks inside the ROI. This can lead to wildly off m/z values when the ROI contains noise.

My use case is that I am targeting certain peaks (by pre-calculating ROIs) for QC purposes and since we had mass stability issues I need to put a pretty wide m/z range to be sure to actually find the peak. But this cases also some tiny noise peaks to be included and the mass returned is then way off.

I could not find the code that actually does this but I calculated by hand that the above is what happens. What I would suggest is to combine the m/z values by intensity weighted mean instead (like it can be done across the peak).
I think this might also be a more gently solution to the orbitrap shoulder issues than the filter we just implemented in Spectra...

jorainer:
What you describe, is this happening during the integration of the detected peak (where the `"mz"`, `"rt"` and `"into"` are calculated)?

The `mzCenterFun` parameter of `CentWaveParam` should actually allow you to define how the m/z of the peak is calculated. The default is `wMean` which *should* be an intensity-weighted mean. Maybe you can test different options (e.g. `"apex"`, `"mean"`) to see what changes?

stanstrup:
`mzCenterFun` defines how the mass is chosen/calculated _between_ the scans. The issues seems to be that if there are multiple m/z peaks _inside_ the same scan (and inside the ROI) it is the simple mean that is used (for the m/z, the intensities are summed), supposedly before any integration.
It is unclear to me when this happens.

An example:
I have a compound with theoretical mass: 522.3554
I create a ROI in a 200 ppm window, i.e. 522.2505 - 522.4595.

In the _apex_ scan I have peaks at:
```
mz            int
522.3558      3e4
522.4548      2e1
```

What I get in the peak table with `mzCenterFun = "apex"` is mz = 522.4053 (= this is the simple mean of the two peaks in the scan). So 95 ppm off even if the main peak is spot on. It is distorted by the small noise peak.
The weighted mean would have been 522.3559 which is just 1 ppm off.

[...]

PR #606, opened 2022-03-25
Fix peak mz

jorainer:
This PR changes the way centWave calculates the m/z of a peak if within the ROI there are more than one mass peak per spectrum. In theory, for well centroided data, this should not happen but can happen sometimes.

The default was to simply calculate the mean m/z across all peaks per spectrum (retention times) within the ROI. This PR changes this and uses a intensity weighted mean instead of the plain mean. This should fix issue #590

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: xcms
-Version: 3.17.4
+Version: 3.17.5
 Title: LC-MS and GC-MS Data Analysis
 Description: Framework for processing and visualization of chromatographically
     separated and single-spectra mass spectral data. Imports from AIA/ANDI NetCDF,
--- a/R/do_findChromPeaks-functions.R
+++ b/R/do_findChromPeaks-functions.R
@@ -374,7 +374,7 @@ do_findChromPeaks_centWave <- function(mz, int, scantime, valsPerSpect,
         idxs <- which(eic$scan %in% seq(scrange[1], scrange[2]))
         mzROI.EIC <- list(scan=eic$scan[idxs], intensity=eic$intensity[idxs])
         ## mzROI.EIC <- rawEIC(object,mzrange=mzrange,scanrange=scrange)
-        omz <- .Call("getMZ", mz, int, scanindex, as.double(mzrange),
+        omz <- .Call("getWeightedMZ", mz, int, scanindex, as.double(mzrange),
                      as.integer(scrange), as.integer(length(scantime)),
                      PACKAGE = 'xcms')
         ## omz <- rawMZ(object,mzrange=mzrange,scanrange=scrange)
@@ -908,7 +908,7 @@ do_findChromPeaks_centWave <- function(mz, int, scantime, valsPerSpect,
         ## original mzROI range
         idxs <- which(eic$scan %in% seq(scrange[1], scrange[2]))
         mzROI.EIC <- list(scan=eic$scan[idxs], intensity=eic$intensity[idxs])
-        omz <- .Call("getMZ", mz, int, scanindex, as.double(mzrange),
+        omz <- .Call("getWeightedMZ", mz, int, scanindex, as.double(mzrange),
                      as.integer(scrange), as.integer(length(scantime)),
                      PACKAGE = 'xcms')
         if (all(omz == 0)) {
--- a/R/do_findChromPeaks-functions.R
+++ b/R/do_findChromPeaks-functions.R
@@ -338,6 +338,9 @@ do_findChromPeaks_centWave <- function(mz, int, scantime, valsPerSpect,
     Nscantime <- length(scantime)
     lf <- length(roiList)
 
+    ## @JO REMOVEME
+    mz_fun <- getOption("MZFUN", default = "getMZ")
+
     ## cat('\n Detecting chromatographic peaks ... \n % finished: ')
     ## lp <- -1
     message("Detecting chromatographic peaks in ", length(roiList),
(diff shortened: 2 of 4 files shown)


jorainer:
@sneumann happy to discuss this further. @stanstrup , would be happy to know if this indeed fixes your issue #590.

sneumann:
I would love to have a script snippet I can run across a number of files, and get a summary (mean, median, quantiles, possibly rmsd) with/without the change. And a graphical plot of the ROIs with both before/after mzs as lines for the "worst" changing ones. I am happy to run that on numerous of our instrument data. Yours, Steffem

jorainer:
I'll update this PR to (temporarily) allow switching between the old and new m/z calculation. Then it should be straight forward to compare/evaluate.

jorainer:
@sneumann I have now added the possibility to choose the m/z calculation function (global option `MZFUN`). An example workflow how the impact can be evaluated:

Load test data and define the cent wave parameters. Note: I've used here `ppm = 100` to actually start seeing an effect on the present data set. The bigger the `ppm` parameter the bigger the impact of the weighted mean is.

```r
library(xcms)
fls <- dir(system.file("sciex", package = "msdata"), full.names = TRUE)[1L]

data <- readMSData(fls, mode = "onDisk")
data <- data |>
    smooth(method = "SavitzkyGolay", halfWindowSize = 6) |>
    pickPeaks()
cwp <- CentWaveParam(peakwidth = c(2, 10), ppm = 100, integrate = 2)
```

Run *default* centWave:

```r
options(MZFUN = "getMZ")
data_old <- findChromPeaks(data, cwp)
```

Use the new weighted average m/z calculation.

```r
options(MZFUN = "getWeightedMZ")
data_new <- findChromPeaks(data, cwp)
```

Identify peaks with the largest impact:

```r
mzdiff <- chromPeaks(data_old)[, "mz"] - chromPeaks(data_new)[, "mz"]

sum(mzdiff != 0)
idx <- head(order(abs(mzdiff), decreasing = TRUE))
```

Compare the effect: plot the peak with type `"XIC"` and indicate the peak's m/z value based on the original calculation (red) and new one (blue).

```r
i <- 1L

a <- data_old |>
filterRt(chromPeaks(data_old)[idx[i], c("rtmin", "rtmax")] + c(-2, 2)) |>
filterMz(chromPeaks(data_old)[idx[i], c("mzmin", "mzmax")] + c(-0.02, 0.02))

plot(a, type = "XIC")
abline(h = chromPeaks(data_old)[idx[i], c("mz")], col = "red")
abline(h = chromPeaks(data_new)[idx[i], c("mz")], col = "blue", lty = 2)
```
A

With the standard mean the m/z of the peak is actually quite off (horizonal red line above). The blue dashed line represents the calculation using the weighted mean (closed to the *real* m/z values). Also, the peak region (red rectangle) seems to be too wide. The same plot for the peak region being defined also depending on the weighted mean:

```r
b <- data_new |>
filterRt(chromPeaks(data_new)[idx[i], c("rtmin", "rtmax")] + c(-2, 2)) |>
filterMz(chromPeaks(data_new)[idx[i], c("mzmin", "mzmax")] + c(-0.02, 0.02))

plot(b, type = "XIC")
abline(h = chromPeaks(data_old)[idx[i], c("mz")], col = "red")
abline(h = chromPeaks(data_new)[idx[i], c("mz")], col = "blue", lty = 2)
```
B

The defined peak regtion is also better in this case.

So, summarizing: I think we should switch to use the weighted mean m/z calculation. For nicely centroided data there should not be big differences, but with less nice centroided data or larger `ppm` the weighted median should yield more accurate results.

stanstrup:
Yes. I am so embarrassed for not getting back to you. I have just been slammed with running a course. But I am looking at it right now. Unfortunately I didn't keep track of my problematic file so I am trying to run an entire analysis on a dataset for the mass range for that example compound. Takes a while pulling the data at home. But I should be able to get back to you in a few days with something. I think your example already shows that it improves results.

Now that I think of it I will also try to find the orbitrap data we used for the artifact filter. I would be interested in seeing if this is a less aggressive solution to that problem too.

stanstrup:
It looks to me like this is a very significant improvement.
Here is an example run at 200 ppm where I have put the theoretical mass of the compound in green:

Old algorithm's windows:
image

New algorithm's windows:
image


The new algorithm gets it spot on and has a much more precise window.


I then checked across the entire dataset and it clearly shows that the new algorithm gets the mass right every time and the old one doesn't.

image


If I lower the ppm to something more normal like 30 ppm then I get practically identical results for this data with old and new algo but it really depends on the noise structure.


I took a look at the orbitrap file we use for the FFT filter example. I needed to increase the ppm to 50 ppm to get a difference for thisd peak but here the old algorithm would be ~40 ppm off due to the noise. I imagine that with the new algorithm you might not need the heavy-handed FFT artifact filter but could do something faster with peak picking following by `filterChromPeaks` (something like group by scpos and mz and keep max(maxo)) before grouping.

image


So as far as I can see the weighted mean:
* makes it more robust to too wide ppm
* makes the mz robust to noise
* if you do something targeted with a predefined wide window it gives more accurate results for the mass

I cannot see a scenario where this is not advantageous.

jorainer:
Great! Thanks Jan for this extensive evaluation! Maybe if @sneumann could also have a look at it we could get this merged to get included into the next BioC release?

sneumann:
Hi, I now got around to check ~17000 mzML files with our normal parameters (QTOF, 25ppm), and found 34 files with differences, everything else unchanged. Interestingly, all of them had just one peak differing, and 25 had a negative difference, and 10 a positive one. So the fix seems to work without major side-effects.

Issue #612, opened 2022-04-13
plot chromatograms - Stacking EICs


CarlBrunius:
Hi guys,
When examining multiple EIC for a particular feature, I find it hard to visualize more than a rather low number of samples. In lots of conventional software, there is frequently an option of stacking multiple spectra in the y axis by adding a pseudo-offset between subsequent injections. 

Let's say I've identified a feature of interest and want to confirm that the peak looks good. I'd do something like this:

```
fd <- featureDefinitions(xdata)

# let's say you're interested in features close to mz=201.113 and rt=222.95
target_mz <- 193.131
target_rt <- 258.7440
whichFt <- which(abs(fd$mzmed - target_mz) < 0.001 & abs(fd$rtmed - target_rt) < 3)
ft1 <- fd[whichFt,]

# Plot feature peaks
par(mfrow = c(2, 1))
feature_chroms <- featureChromatograms(xdata, features = whichFt, expandRt = 20, adjustedRtime = F)
plot(feature_chroms, peakType = 'rectangle', peakBg = NA)
feature_chroms <- featureChromatograms(xdata, features = whichFt, expandRt = 20, adjustedRtime = T)
plot(feature_chroms, peakType = 'rectangle', peakBg = NA)

```
![image](https://user-images.githubusercontent.com/12426000/163234926-42d3d8fe-2ad8-49ad-8e11-e8e4d3b86be7.png)
In this particular example, the peaks look really good. In fact, it looks like the retention adjustment is even making RTs more dissimilar in this particular region :)

But if I'm in a more crowded region with not too good retention time adjustment, or potentially in a region with a lot of isomers, peaks may be a little all over the place. It would be awesome to see them stacked to get a better overview of peaks in the individual injections.

Moreover, the options for expanding mz and rt with custom ranges would be awesome also for feature peaks.

To customize a bit more I tried to do this:

```
# Extract the chromatographic peak numbers corresponding to your Feature of Interest (FoI)
ft1_peaks <- ft1$peakidx[[1]]

# Extract all chromatographic peaks
cp <- chromPeaks(xdata)

# Restrict chromatographic peaks to the FoI
cp1 <- cp[ft1_peaks,]

# Extract chromatograms
mzr <- target_mz + c(-0.005, 0.005)
rtr <- target_rt + c(-40, 40)
par(mfrow = c(2, 1))
chr <- chromatogram(xdata, mz = mzr, rt = rtr, adjustedRtime = F)
[...]

PR #614, opened 2022-04-14
feat: add parameter expandMz to featureChromatograms

jorainer:
- Add parameter `expandMz` to `featureChromatorgrams` (#612).
- Add package `signal` to Suggests and to the direct injection vignette to ensure the package is installed (seems to be now a dependency of the `CWT` package.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: xcms
-Version: 3.17.4
+Version: 3.17.5
 Title: LC-MS and GC-MS Data Analysis
 Description: Framework for processing and visualization of chromatographically
     separated and single-spectra mass spectral data. Imports from AIA/ANDI NetCDF,
@@ -70,7 +70,8 @@ Suggests:
     pheatmap,
     Spectra (>= 1.1.17),
     MsBackendMgf,
-    progress
+    progress,
+    signal
 Enhances:
     Rgraphviz,
     rgl,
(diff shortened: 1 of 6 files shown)


jorainer:
Might be maybe also nice to get this PR in before feature freeze @sneumann? - otherwise we do it after the BioC release.

R/functions-utils.R
```diff
@@ -604,16 +604,16 @@ rowRla <- function(x, group, log.transform = TRUE) {
...
+.update_feature_definitions <- function(x, original_names, subset_names) {
     ## Skip if they are the same.
     if (length(original_names) == length(subset_names) &&
         all.equal(original_names, subset_names))
         return(x)
-    x$peakidx <- bplapply(x$peakidx, function(z) {
-        idx <- base::match(original_names[z], subset_names)
-        idx[!is.na(idx)]
-    }, BPPARAM = BPPARAM)
+    f <- as.factor(rep(seq_len(length(x$peakidx)), lengths(x$peakidx)))
```
sneumann, line 612: Rationale here was that `bpapply` had too much overhead for that simple task ?

jorainer: `bplapply` (and also `lapply`) was just very slow, not really because of the parallel processing overlap but more because a `match` call against all names of chromatographic peaks was performed for **every** feature. This version is now vectorized performing the `match` in one go. Calling e.g. `filterFile` with `keepFeatures = TRUE` on a data set with ~ 5000 features across 10 files took ~ 30 seconds with the old version and now it takes less than a second. The performance gain is expected to increase with data set size.

PR #654, opened 2023-01-14
Change rgl.* to *3d

dmurdoch:
Some upcoming changes to the rgl package will require changes to
xcms.  I will be deprecating a number of rgl.* function
calls. You can read about the issues here:

             https://dmurdoch.github.io/rgl/articles/deprecation.html

I have made the required changes to xcms in the attached
patch file. These changes should work with both old and new rgl
versions, though some default colors and projections have changed.

Duncan Murdoch


jorainer:
Thanks @dmurdoch ! I will merge (and fix then the unit tests)

PR #666, opened 2023-03-10
xcms 4: native support for MsExperiment and Spectra

jorainer:
This PR adds native support for `Spectra` to `xcms`: in addition to a `OnDiskMSnExp` it is now possible to run the full `xcms` pre-processing on a `MsExperiment` object that bases/uses `Spectra` for MS data representation.

Also, some internal functions were refactored which should result in some significant speed improvement for some tasks.


R/XcmsExperiment.R
```diff
@@ -0,0 +1,1707 @@
+#' @title Next Generation `xcms` Result Object
```
sneumann, line 1: Cosmetic hint on framing: "next generation" rarely ages well. Similar to "-final" in word files, where inevitably "-final2" crops up.

tests/testthat/test_XcmsExperiment-functions.R
```diff
@@ -0,0 +1,125 @@
...
+
+    ## also check that it works on `XcmsExperiment`.
+    mzc <- MzClustParam(sampleGroups = c(1, 1))
+    tmp <- groupChromPeaks(tmp, param = mzc)
+    cpp <- ChromPeakAreaParam()
+    res <- fillChromPeaks(tmp, cpp)
+    expect_true(length(res@processHistory) > length(tmp@processHistory))
+    expect_true(sum(is.na(featureValues(res))) < sum(is.na(featureValues(tmp))))
+
+    ## Compare with XCMSnExp.
```
sneumann, line 119: I like that idea to cross-compare results. I guess many of those implicit checks now are being checked more explicitly

sneumann:
Hi, I am super-impressed by the amount of work into this branch, and looking forward to use the XcmsExperiment based classes in the future ! Yours, Steffen

R/XcmsExperiment.R
jorainer: good point - do you have an alternative for a good title?

jorainer:
as agreed I closed this PR and will directly push to devel

PR #667, opened 2023-03-13
tests: fix longtests

jorainer:
Small fixes to the longtests and fix for the CI GHA.


sneumann (approved):
Excellent, works like a charm here!

Issue #693, opened 2023-10-12
Properly serialize an xcms result object to disk/files


jorainer:
Instead of exporting/saving an `xcms` object in RData format it would be good to also support import/export from/to files in textual format.

Idea:
- re-create an xcms result object from plain text files.
- export the xcms result to text files.

Why? mostly for Galaxy-based workflows, to enable usage of the result objects across tools.

jorainer:
@hechth

sneumann:
How much of that could be handled by mzTab-M ? The SMF Feature table nicely corresponds to the (grouped) xcms list. Ungrouped data is technically also kinda possible and requires lots of repetition. Metadata works as well. Bonus: we'd be able to read results from e.g. MS-Dial for further analysis. 

Alternatively there are some functions by @PayamEmami in the MetaboIgniter workflow
https://github.com/nf-core/metaboigniter/blob/master/bin/consensusXMLToXcms.r
https://github.com/nf-core/metaboigniter/blob/master/bin/featureXMLToXcms.r

An earlier version was even implemented in Galaxy to integrate OpenMS and R based tools.

Main question would be if true round-tripping (export xcms object1 and re-import to object2 guaranteeing object1==object2) is a requirement.  

Yours, Steffen

jorainer:
Agree that it would be great to use standard file formats as much as possible! maybe we could export the feature data as a mzTab-M and the rest as other files (all in one folder). I would like to be able to export a full xcms result object to completely restore it again later (including process history, adjusted retention times, etc).

jorainer:
@sneumann , do you have some definition somewhere how the mzTab-M format should actually look like?

sneumann:
Hi, the official docs are in 
https://hupo-psi.github.io/mzTab/2_0-metabolomics-release/mzTab_format_specification_2_0-M_release.html
and @nilshoffmann wrote an R package to read/write mzTab-M at https://github.com/lifs-tools/rmzTab-m/
I need to find some snippets I have to export the old xcms stuff which could become a starting point, 
and we'd need to figure how to do the dependencies. Does that sound like an MsBackendMzTabM ? 
Yours, 
Steffen

jorainer:
Had a quick look at the `rmzTab-m` - but am not particularly happy with the R6 objects. Would want to avoid dependency of `xcms` on such objects as much as possible.

sneumann:
+1 for not having all that as dependency. 
[...]

PR #698, opened 2023-10-26
feat: creation of new generic function so save XcmsExperiment object

philouail:
Hi so this is just a draft, but  I thought it might be easier to discuss here. I might need a bit of back and forth as it's the first time I do that :) Hope it's fine for you to do it like that. I'll also leave it as draft to avoid confusion. 

See my comments for the parts that I'm confused about !


R/AllGenerics.R
```diff
@@ -1943,6 +1943,51 @@ setGeneric("stitch", function(object, lockMass, ...) standardGeneric("stitch"))
...
+#'
+#' The `storeResults` function saves an `object` resulting from processing with
+#' the `xcms` package (mainly `XcmsExperiment`). Multiple formats for storing
+#' and exporting are available and can be defined by the `param` argument.
+#'
+#' Supported `param` objects are:
+#' 
+#' - `RDataParam`: Save in an .RData format file.
+#' 
+#' - `PlainTextParam`: Save in a plain text format (to be defined).
```
philouail, line 1960: Is it fine to add the parameters that are not defined yet ?

R/RDataParam.R
```diff
@@ -0,0 +1,70 @@
+#' @title Store `XcmsExperiment` object as .RData file
+#'
```
philouail, line 2: in general i'm a bit confused by what documentation a single method/new class is supposed to have. I have not found an example similar to this.  I have mostly found generic function detailed documentation. So I'm not sure about the structure and level of detail here.

tests/testthat/test_XcmsExperiment.R
```diff
@@ -1326,3 +1315,11 @@ test_that("setAs,XcmsExperiment,xcmsSet works", {
...
     expect_equal(peaks(res), chromPeaks(xmseg))
 })
+
+test_that("storeResults,RDataParam works", {
+    param <- RDataParam(fileName ="test")
+    param2 <- RDataParam()
+    expect_false(is.null(param2))
+    storeResults(xmse, param = param)
+    expect_true(file.exists("test"))
+})
```
philouail, line 1328: Do you think about any other ? 
I was thinking i could also try to load() it and see its class, but maybe this is overkill.

R/RDataParam.R
```diff
@@ -0,0 +1,70 @@
...
+#' .RData file.
+#' 
+#' @return The saved object as an .RData file. 
+#'
+#' @author Philippine Louail
+#'
+#' @examples
+#'
+#'## Get an `XcmsExperiment` object
+#' x <- ...
```
philouail: Not sure where to get an XcmsExperiment object, should I do something similar to the testthat.R file ? But then I would need to load packages and i don't know if that's okay.

tests/testthat/test_XcmsExperiment.R
```diff
@@ -633,7 +622,7 @@ test_that(".xmse_filter_peaks_intensities works", {
     res <- .xmse_filter_peaks_intensities(xmse, nValues = 1, threshold = 50000)
     expect_equal(res, unname(chromPeaks(xmse)[, "maxo"] >= 50000))
 
-    res <- .xmse_filter_peaks_intensities(xmse, nValues = 1, , msLeve = 2L)
+    res <- .xmse_filter_peaks_intensities(xmse, nValues = 1, , msLevel = 2L)
```
philouail: Completely unrelated I found this typo aha

philouail:
Oh and also I tried to do the necessary changes in `description` and `NEWS` files, but can you check if it looks OK ? Maybe there's different rules for XCMS compared to the other packages I contributed. 
And I have to look into this NAMESPACE thing, to see how to do it manually tomorrow.

R/AllGenerics.R
jorainer: Yes, that's fine. But please use `[RDataParam()]` instead of `RDataParam` - roxygen will then add a link to the RDataParam documentation (note: you can't that do that now for the `PlainTextParam` because documentation does not yet exist ;) ).

R/RDataParam.R
jorainer: The general documentation in `AllGenerics` should basically tell general information, i.e. that these function allow to export/import results in various formats and that the format depends on the used param objects. In the documentation here (for the parameter object) you should provide information about the format in which the result is stored, i.e. here basically just say that the object gets exported using the `[save()]` function and imported using `[load()]`. I don't think we would need to provide much more information here. You would then also need to provide documentation for the different parameters of the param object.

For the text file format export we will need more information, like a listing of all files that will be exported, its format etc. So basically describe that then more in detail to eventually allow also other developers (e.g. from Python) to understand how such results could be read and what they mean.

R/AllGenerics.R
```diff
@@ -1943,6 +1943,51 @@ setGeneric("stitch", function(object, lockMass, ...) standardGeneric("stitch"))
 setGeneric("stitch.xml", function(object, lockMass) standardGeneric("stitch.xml"))
 setGeneric("stitch.netCDF", function(object, lockMass) standardGeneric("stitch.netCDF"))
 setGeneric("stitch.netCDF.new", function(object, lockMass) standardGeneric("stitch.netCDF.new"))
+
+#' @title Save Xcms objects in a specified format
```
jorainer: I would maybe say here xcms result objects.

R/AllGenerics.R
```diff
@@ -1943,6 +1943,51 @@ setGeneric("stitch", function(object, lockMass, ...) standardGeneric("stitch"))
...
+#'
+#' @param param The parameter object selecting and configuring the format for
+#' saving. It can be one of the following classes: `RDataParam`,
+#' `PlainTextParam`, or `MzTabMParam`.
+#'
+#' @param ... Optional parameters.
+#'
+#' @name storeResults
+#' 
+#' @author Philippine Louail, Johannes Rainer
```
jorainer: you can drop my name here :) I did not provide code.

R/RDataParam.R
jorainer: you can use this:

```
#' ## Load a test data set with detected peaks
#' faahko_sub <- loadXcmsData("faahko_sub2")

```

tests/testthat/test_XcmsExperiment.R
jorainer: thanks! :)

tests/testthat/test_XcmsExperiment.R
jorainer: Better to test more than less. I would actually load it again and see if the objects are the same.

jorainer:
For the NAMESPACE, just add 

```
export RDataParam
exportMethods storeResults
```

that should be enough. Description and NEWS are OK - yes, there is a different format here, but what you added is fine.

R/RDataParam.R
philouail: Ok So I remove everything and just leave @title, @name, @description and the 2 @param ? Or I keep it as it is ? 
I will add a sentence that specify that it gets exported using the `[save()]`.

jorainer: Actually, you will need to add also a `@param fileName` (and all other parameters for `RDataParam`). I would however remove the *Other available `param` classes* part. For `@return` I would write:

```
@return for `RDataParam`: a `RDataParam` class. `storeResults` does not return anything but saves the object to a *RData* file.
```

jorainer: And to link to the other parameters/options you can add a

```
@family xcms result export formats
```

somewhere in the documentation. the same should then also be added for the `PlainTextParam` and roxygen will then take care of linking the two together.

philouail:
Hi @sneumann so this PR is related to #693, it establish a new generics function `storeResults`  that will allow to save xcms results objects into different format. This PR also implement the first format which is .RData with the `RDataParam`. 
I will then look into the Plain text file format and lastly MzTab-m they will each have they own param class and specific method. 

Tell me if there is anything unclear or I need to change something.

src/xcms.dll
jorainer, line 1: can you please delete that file? It's anyway strange that it shows up/is added by git. Maybe add `*.dll` to the `.gitignore` file of xcms.

jorainer (approved):
Thanks Phili! I'm fine with this PR. 

Please @sneumann also have a look whether you're OK with the general approach we take here.

Only thing, @philouail , please address the one point regarding the dll file.

src/xcms.dll
philouail: yes not sure what happened here, but removed it and added the extension to `.gitignore` file !

tests/testthat/test_XcmsExperiment.R
sneumann: Is the `, ,` on purpose ?

sneumann (approved):
Hi, looks good to me, thanks for the nice work ! 
Yours, Steffen

tests/testthat/test_XcmsExperiment.R
philouail: It is not my code, i think there was some sort of error when copy/pasting. @jorainer Is it fine that I remove the extra comma ?

jorainer: hm, thanks for the catch @sneumann . Yes @philouail , please remove the extra comma.

sneumann:
Cool, so ready to merge from my side.

PR #699, opened 2023-11-14
Addition of PlainTextParam to the storeResults generic function

philouail:
This PR is for the addition of a method to save `XcmsExperiment` and `MsExperiment` objects as a collection of plaint text files.

The description of the Param details explains the name of the text files saved and what they contain/what function was operated to get the contents

@jorainer , here it is with the small changes that we talked about. 
One thing, I am not sure what action is added to the processing queue exactly and it's blocking me to do one the of UT (see my comments there). 
Also I need and `XcmsExperiment` object with features definitions, chrompeaks data, features definitions and all for the UT(you'll see it failed for the signature with `XcmsExperiment`). There is none in the `testthat.R` should I create one or there is something else I can use ?


R/PlainTextParam.R
```diff
@@ -0,0 +1,215 @@
+#' @title Store `MsExperiment` and `XcmsExperiment` object as a folder of plain
```
jorainer: Maybe alternative title:

```
@title Store contents of `MsExperiment` and `XcmsExperiment` objects as plain text files
```

I don't think we need to specify in the title that these files will be placed to a specified folder.

R/PlainTextParam.R
```diff
@@ -0,0 +1,215 @@
...
+#' `XcmsExperiment` objects as collections of plain text files in a specified
+#' folder.
+#'
+#' For an `MsExperiment` object, the exported files include:
+#'
+#' - The [`sampleData()`] stored as a text file named "sample_data.txt".
+#'
+#' - The [`fileNames()`] of the *Spectra* object stored in a tabular format in a
+#' text file named "spectra_files.txt". This method only works if the `Spectra`
+#' object has an backend `MsBackendMzR`. If it does not, not spectra data will
```
jorainer: this part of the sentence is a bit odd. maybe write (instead of `This method...` and the following sentence):

```
The file names will only be exported if the `Spectra` object uses a [MsBackendMzR()] backend. 
For other backends no information on raw spectra data is currently exported with `PlainTextParam`.

```

R/PlainTextParam.R
```diff
@@ -0,0 +1,215 @@
...
+#' @name PlainTextParam
+#'
+#' @export
+#'
+#' @family xcms result export formats.
+#'
+#' @description
+#' The `PlainTextParam` class and method enable users to save `MsExperiment` or
+#' `XcmsExperiment` objects as collections of plain text files in a specified
+#' folder.
```
jorainer: Maybe also add here a note that exporting `MsExperiment` currently only works with `MsBackendMzR`? Something like:

```
... folder. Note that, while for all xcms results within the `XcmsExperiment` can and will be exported, 
the full raw MS data (of the object's `Spectra` object) will currently not be exported in plain text format. 
For `Spectra` using the [MsBackendMzR()] backend, the names of the raw data files will however be 
exported (which enables to *restore* the full `Spectra` respectively `MsExperiment` objects).
```

R/PlainTextParam.R
```diff
@@ -0,0 +1,215 @@
...
+#'
+#' For an `MsExperiment` object, the exported files include:
+#'
+#' - The [`sampleData()`] stored as a text file named "sample_data.txt".
```
jorainer: can you maybe use `"` for the file names instead of only "? Or, alternatively * (the filenames would then be rendered in italics font)?

jorainer: With `"` I meant a " with the backticks - but maybe using italics for file names would be better?

R/PlainTextParam.R
```diff
@@ -0,0 +1,215 @@
...
+#' - The [`featureDefinitions()`] stored in a text file named
+#' "feature_definitions.txt". Additionally, a second file named
+#' "feature_peak_index.txt" is generated to connect the features' definitions
+#' with their names.
+#'
+#' This `param` class and method are part of the possible dispatch of the
+#' generic function `storeResults`.
+#'
+#' @param path for `PlainTextParam` `character(1)`, defining where the files are
+#' going to be stored. The default will be `tempdir()`.
```
jorainer, line 57: Also maybe add:

```
The folder will be created by calling `storeResults`. If the folder already exists, previous exports 
in that folder might get overwritten.
```

tests/testthat/test_XcmsExperiment.R
```diff
@@ -1326,3 +1326,34 @@ test_that("storeResults,RDataParam works", {
     expect_s4_class(object, "XcmsExperiment")
     expect_equal(object, xmse)
 })
+
+test_that("storeResults,PlainTextParam,MsExperiment works", {
+    param <- PlainTextParam(path = "/path/test/")
```
jorainer: just to be on the safe side - also seeing that unit tests fail on the github actions: can you use here a path:

```
pth <- file.path(tempdir(), "test")
param <- PlainTextParam(path = pth)
```

tests/testthat/test_XcmsExperiment.R
```diff
@@ -1326,3 +1326,34 @@ test_that("storeResults,RDataParam works", {
...
+    param2 <- PlainTextParam()
+    expect_false(is.null(param2))
+    # Test if PlainTextParam is invalid with multiple paths
+    expect_error(new("PlainTextParam", path = c(tempdir(), tempdir())))
+    mse <- filterRt(mse, c(100,150)) #this adds to processing but not processingQueue
+    storeResults(mse, param = param)
+    expect_true(dir.exists("/path/test/"))
```
jorainer: and then here 

```
expect_true(dir.exists(pth))
```

tests/testthat/test_XcmsExperiment.R
jorainer: you will in fact get problems storing directly to the root directory `/` because of permissions.

tests/testthat/test_XcmsExperiment.R
```diff
@@ -1326,3 +1326,34 @@ test_that("storeResults,RDataParam works", {
...
     expect_equal(object, xmse)
 })
+
+test_that("storeResults,PlainTextParam,MsExperiment works", {
+    param <- PlainTextParam(path = "/path/test/")
+    param2 <- PlainTextParam()
+    expect_false(is.null(param2))
+    # Test if PlainTextParam is invalid with multiple paths
+    expect_error(new("PlainTextParam", path = c(tempdir(), tempdir())))
+    mse <- filterRt(mse, c(100,150)) #this adds to processing but not processingQueue
```
jorainer: filtering on retention time might actually already do the filtering and not put that into the processing queue (because it is the same as just subsetting the `Spectra`). Operations performed on peak data will however be put into the processing queue, so better to call e.g. `mse <- filterMzRange(mse, c(200, 500))`.

tests/testthat/test_XcmsExperiment.R
```diff
@@ -1326,3 +1326,34 @@ test_that("storeResults,RDataParam works", {
...
+    storeResults(mse, param = param)
+    expect_true(dir.exists("/path/test/"))
+    expect_true(file.exists(file.path(param@path, "sample_data.txt")))
+    expect_true(file.exists(file.path(param@path, "spectra_files.txt")))
+    expect_true(file.exists(file.path(param@path, "spectra_processing_queue.json"))) # empty so does not work
```
jorainer: A `filterMzRange` call will show up in the processing queue (while `filterRt` will not) - so, with that this file should actually be present then.

tests/testthat/test_XcmsExperiment.R
```diff
@@ -1326,3 +1326,34 @@ test_that("storeResults,RDataParam works", {
...
+})
+
+test_that("storeResults,PlainTextParam,XcmsExperiment works", {
+    param <- PlainTextParam(path = "/path/test/")
```
jorainer: Also here, just to be on the safe side, use `path = file.path(tempdir(), "test")`

jorainer:
I would suggest to move the unit tests you have now defined into the *tests/test_PlainTextParam.R* file (then all the code in the *PlainTextParam.R* would have the unit tests in that file).

To get a `XcmsExperiment` object for your unit test you can use the `loadXcmsData` function to load one of the bundled (full) result objects. Maybe put something like this at the beginning of your unit test file:

```r
xmse_full <- loadXcmsData("xmse")
```

I would avoid to use the variable `xmse` again, because it could then overwrite the one created in *testthat.R* and some unit tests might mysteriously fail afterwards.

That `XcmsExperiment` has all chrom peaks, adjusted retention times and correspondence results, so you can then check for presence of all files. To check the `storeResults,MsExperiment,PlainTextParam` method you can still use the `mse` variable, as that one is created in *testthat.R*.

jorainer (changes requested):
Thanks Phili! Great PR!

I requested some changes, mostly in the documentation. Also, I would suggest to move the unit tests into the *test_PlainTextParam.R* file - and use one of the bundled xcms result objects (see comment in the *Conversation*).

R/PlainTextParam.R
philouail: I like italics, i will go with that

tests/testthat/test_XcmsExperiment.R
philouail: Ok good to know ! thanks.

philouail:
just to be sure the *test_PlainTestPram.R* file should be in the *testthat/* folder right ?

jorainer:
> just to be sure the test_PlainTestPram.R file should be in the testthat/ folder right ?

Yes, exactly.

jorainer (approved):
Great contribution, thanks Phili! I'm OK with the PR.

jorainer:
@sneumann can you please also check this PR? If OK, please merge and push to Bioconductor.

sneumann:
Hi, PR looks good to me. So currently this is write-Only, we don't have any consumer of the JSON, 
and don't use the `read_json()` to read back. Is that correct ? Is this next on the list ? Yours, Steffen

sneumann (approved):
Thanks for the contribution ! Yours, Steffen

jorainer:
Hi Steffen @sneumann , yes, this is currently only write-only. We wanted to first have export functionality (including mzTab-M which will come in a separate package) and once we're OK with it add import/read functions. And yes, we re-import using `read_json`.

Issue #702, opened 2023-11-29
Definition of anchor peaks for subset-based peakGroups alignment not correct 


jorainer:
The current implementation of the subset-based peakGroups alignment to define the anchor peaks currently does not correctly select the peaks based on the `minFraction` parameter: `minFraction` is referring to the total number of samples and not the total number of samples **of the subset**. Example:

```r
library(xcms)
faahko_3_files <- c(system.file('cdf/KO/ko15.CDF', package = "faahKO"),
                    system.file('cdf/KO/ko16.CDF', package = "faahKO"),
                    system.file('cdf/KO/ko18.CDF', package = "faahKO"))

faahko_od <- readMSData(faahko_3_files, mode = "onDisk")
xod_x <- findChromPeaks(
    faahko_od, param = CentWaveParam(noise = 10000, snthresh = 40,
                                     prefilter = c(3, 10000)))
pdp <- PeakDensityParam(sampleGroups = rep(1, 3))
xod_xg <- groupChromPeaks(xod_x, param = pdp)
```

Using the first and last sample for subset and setting `minFraction = 0.6` we would expect to get anchor peaks only if a peak was identified in both samples, but

```r
p <- PeakGroupsParam(subset = c(1, 3), minFraction = 0.6)
adjustRtimePeakGroups(xod_xg, p)
     ko15.CDF ko18.CDF
FT21 2679.783       NA
FT15 2678.218 2684.479
FT16 2679.783 2682.914
FT20 2679.783 2684.479
FT24 2681.348 2684.479
FT08 2783.070       NA
FT04 2786.200 2787.766
FT46 2786.200 2787.766
FT01 2787.765 2787.766
FT45       NA 2916.092
FT06 2923.916 2923.917
FT05 2994.338 2994.339
FT02       NA 3250.992
FT27 3276.030 3285.421
FT31 3316.719 3340.194
FT41       NA 3349.584
FT03 3379.317 3387.143
FT32 3384.012 3410.617
FT33       NA 3410.617
FT11 3413.746       NA
FT10 3415.311 3416.877
FT28 3454.435       NA
FT42 3476.344 3482.605
FT17 3491.994       NA
FT09 3496.688       NA
FT07 3496.688 3501.384
FT35 3512.338 3523.294
FT34 3515.468 3524.859
FT43 3576.501 3584.327
FT30 3590.586       NA
FT18 3610.930 3614.061
FT12 3618.755 3626.581
FT13 3618.755 3626.581
[...]

PR #703, opened 2023-11-29
Fix issue with subset-based peakGroups alignment and other updates

jorainer:
This fixes issue #702 and in addition splits the peakGroups alignment function into two parts (finally) supporting also the use of externally defined alignment matrices.


jorainer:
Note: unit tests fail until Spectra is updates.

R/XcmsExperiment.R
```diff
@@ -1332,35 +1351,28 @@ setMethod(
...
-            subset = subset(param),
-            subsetAdjust = subsetAdjust(param)
-        )
+        rt_adj <- .adjustRtime_peakGroupsMatrix(
+            rt_raw, peakGroupsMatrix(param), smooth = smooth(param),
+            span = span(param), family = family(param),
+            subset = subset(param), subsetAdjust = subsetAdjust(param))
         pt <- vapply(object@processHistory, processType, character(1))
         idx_pg <- .match_last(.PROCSTEP.PEAK.GROUPING, pt, nomatch = -1L)
         if (idx_pg > 0)
```
philouail, line 1380: the double entry has to happen here I guess @jorainer

jorainer: actually, it was the `.match_last` call. Instead of returning `-1L` if no peak grouping was performed it returned a positive integer (I think it was 3 in that particular case), so, subset the processHistory with that "out of bound" index returned/added the `NULL` entry. Stupid but oin the `.match_last` function.

jorainer:
OK. unit tests fixed - ready for review/comments @sneumann

R/MsExperiment.R
```diff
@@ -20,6 +20,14 @@ setMethod("filterMz", "MsExperiment",
               filterMzRange(object, mz, msLevel.)
           })
 
+#' @rdname XcmsExperiment
+setMethod("filterMsLevel", "MsExperiment",
+          function(object, msLevel. = uniqueMsLevels(object)) {
```
sneumann, line 25: First time I see a parameter with trailing dot. Is there a pattern I need to learn ?

R/do_adjustRtime-functions.R
```diff
@@ -115,282 +121,184 @@ do_adjustRtime_peakGroups <-
...
-        ## Finally applying the correction
-        rtime[[i]] <- rtime[[i]] - rtdevsmo[[i]]
-    }
-
-    if (warn.overcorrect) {
-        warning("Fitted retention time deviation curves exceed points by more",
-                " than 2x. This is dangerous and the algorithm is probably ",
-                "overcorrecting your data. Consider increasing the span ",
-                "parameter or switching to the linear smoothing method.")
+        rtime_adj
```
sneumann, line 299: I like a more explicit return(rtime_adj), unless we use a style convention that suggests brevity rulez

R/do_adjustRtime-functions.R
```diff
@@ -485,6 +393,10 @@ do_adjustRtime_peakGroups_orig <- function(peaks, peakIndex, rtime,
 #' @details This function is called internally by the
 #'     do_adjustRtime_peakGroups function and the retcor.peakgroups method.
 #'
+#' Update for version 4.1.2: correctly consider the `sampleIndex` and
```
sneumann: I think 4.1.4

data/faahko_sub2.RData
sneumann, line 1: Shouldn't that be in faahKO ?

data/xmse.RData
sneumann, line 1: And this would go into msdata ?

tests/testthat/test_Chromatogram.R
```diff
@@ -1,5 +1,5 @@
 ## test_that("extractChromatograms is deprecated", {
```
sneumann: Not only deprecated, but gone. So test can go out of the window as well

sneumann (commented):
Mostly cosmetics

R/MsExperiment.R
jorainer: Yes, that's rather unusual, I agree. It's still from *MSnbase* days, without the `.` R was complaining about recursive arguments - I guess that came from the R dispatcher that got confused by a parameter `msLevel` and a `msLevel` method. For consistency and backward compatibility we used this also in other method definitions in `Spectra`. Thanks to the partial parameter matching from R users can use `msLevel = 1L` as well as `msLevel. = 1L`.

I know it looks ugly, but we can't change it here, because that comes from far upstream...

data/faahko_sub2.RData
jorainer: Good question. This basically is "just" a small object to have simpler examples (i.e. load the data without having to first perform all preprocessing to then just run the example). 

IMHO the faahKO should "only" contain the raw data and a object of a class from `xcms` should be in `xcms` - that way `faahKO` is independent of `xcms` and we avoid eventual cyclic redundancies.

data/xmse.RData
jorainer: Same here, IMHO `msdata` should contain raw data files and be independent of data structures.

R/do_adjustRtime-functions.R
jorainer: I also used `return()` a lot - but then had a chat with Martin (Morgan) and he mentioned that for R there's no need to have yet another function call at the end of a function since the last variable will anyway be returned by default. `return` should only be used e.g. to break within a function and still return something. That seems to be now the R coding style.

R/do_adjustRtime-functions.R
jorainer: yes sure, thanks! will fix.

tests/testthat/test_Chromatogram.R
jorainer: yep - will remove.

Issue #707, opened 2023-12-14
negative subsetting removes chromPeaks 


philouail:
As the title says, negative subsetting seems to remove the chromPeaks detection even though the default for subsetting is supposed to keep the chromPeaks. 

```r
> test <- loadXcmsData() 
> chromPeaks(test) 
+     head()          
 mz mzmin mzmax       rt    rtmin    rtmax     into     intb  maxo sn sample 
CP0001 594.0 594.0 594.0 2606.910 2586.566 2643.014 161042.2 146073.3  7850 11      1 
CP0002 577.0 577.0 577.0 2610.040 2586.566 2632.005 136105.2 128067.9  6215 11      1 
CP0003 307.0 307.0 307.0 2624.135 2597.520 2650.875 284782.4 264907.0 16872 20      1
CP0004 302.0 302.0 302.0 2622.560 2600.650 2646.158 687146.6 669778.1 30552 43      1 
CP0005 370.1 370.1 370.1 2679.209 2649.303 2706.052 449284.6 417225.3 25672 17      1
CP0006 427.0 427.0 427.0 2680.787 2649.303 2690.263 283334.7 263943.2 11025 13      1 

> test <- test[-1] 
> chromPeaks(test) 
+     head()     
mz mzmin mzmax rt rtmin rtmax into intb maxo sn sample
```

PR #709, opened 2023-12-20
Updates and fixes

jorainer:
- Improve performance of the `chromatogram` call for `XcmsExperiment` objects.
- Remove internal (not exported) normalization functions. These have been transferred to the MetaboCoreUtils package (https://github.com/rformassspectrometry/MetaboCoreUtils/pull/78).
- Support subsetting of `XcmsExperiment` with negative indices (#707).


R/XcmsExperiment-functions.R
```diff
@@ -878,40 +883,49 @@
...
+                                           ":elapsed"),
+                           total = nrow(chrs) + 1L, clear = FALSE)
     for (i in seq_len(nrow(chrs))) {
-        pks <- chromPeaks(object, rt = fd[i, rtc], mz = fd[i, mzc],
-                          msLevel = chrs[i, 1]@msLevel, type = chromPeaks)
+        idx <- .index_chrom_peaks(
+            object, rt = fd[i, rtc], mz = fd[i, mzc],
+            msLevel = chrs[i, 1]@msLevel, type = chromPeaks)
+        f_s <- factor(.chromPeaks(object)[idx, "sample"], levels = js)
+        pkl <- split.data.frame(.chromPeaks(object)[idx, , drop = FALSE], f_s)
```
jorainer, line 899: splitting the chrom peaks `matrix` and chrom peak data `data.frame` considerably improved performance. Also, we try to use `data.frame` instead of `DataFrame` for `chromPeakData` as much as possible: subset of the former is much faster.

R/XcmsExperiment-functions.R
```diff
@@ -878,40 +883,49 @@
...
+    else cpd <- as.data.frame(chromPeakData(object))
+    message("Processing chromatographic peaks")
+    pb <- progress_bar$new(format = paste0("[:bar] :current/:",
+                                           "total (:percent) in ",
+                                           ":elapsed"),
+                           total = nrow(chrs) + 1L, clear = FALSE)
     for (i in seq_len(nrow(chrs))) {
-        pks <- chromPeaks(object, rt = fd[i, rtc], mz = fd[i, mzc],
-                          msLevel = chrs[i, 1]@msLevel, type = chromPeaks)
+        idx <- .index_chrom_peaks(
```
jorainer, line 895: instead of getting the filtered `chromPeaks` matrix we simply get the indices of the rows matching the filter. That allows to more efficiently subset also the `chromPeakData`.

R/XcmsExperiment-functions.R
```diff
@@ -1016,3 +1030,71 @@ featureArea <- function(object, mzmin = min, mzmax = max, rtmin = min,
...
+#' helper function to get indices of chromatographic peaks for eventual
+#' subsetting with `rt`, `mz`, `msLevel` and `type`.
+#'
+#' @param object `XcmsExperiment` or `XCMSnExp` object with `chromPeaks`.
+#'
+#' @return `integer` vector with the indices of the `chromPeaks` matching the
+#'     requested filtering.
+#'
+#' @noRd
+.index_chrom_peaks <- function(object, rt = numeric(),
```
jorainer, line 1043: this function is then also re-used in `chromPeaks,XcmsExperiment` and `chromPeaks,XCMSnExp`.

R/XcmsExperiment-functions.R
```diff
@@ -1016,3 +1030,71 @@ featureArea <- function(object, mzmin = min, mzmax = max, rtmin = min,
...
+            keep <- keep & pks[, "mz"] >= mz[1L] & pks[, "mz"] <= mz[2L]
+    }
+    which(keep)
+}
+
+#' Helper function to return the chromPeakData as-is (as a data.frame) from
+#' a `XcmsExperiment` object.
+#'
+#' @noRd
+.chromPeakData <- function(object, msLevel = integer()) {
```
jorainer, line 1086: Little helper functions to avoid unnecessary conversion of `data.frame` to `DataFrame` for internal calls.

R/XcmsExperiment-functions.R
```diff
@@ -1016,3 +1030,71 @@ featureArea <- function(object, mzmin = min, mzmax = max, rtmin = min,
...
+    if (length(msLevel))
+        object@chromPeakData[object@chromPeakData$ms_level %in% msLevel, ]
+    else object@chromPeakData
+}
+
+#' Helper function to return the **full**  chromPeaks matrix without any
+#' filtering from either a `XcmsExperiment` or `XCMSnExp` object
+#'
+#' @noRd
+.chromPeaks <- function(object) {
```
jorainer, line 1096: That function was required, because we would end up in (infinitely) recursive call with the `chromPeaks` method: `chromPeaks` uses `.index_chrom_peaks`, which in turn calls `chromPeaks` (that again calls `.index_chrom_peaks`, ...). To extract the **full** chrom peaks matrix we can use `.chromPeaks`, to get eventually filtered chrom peaks we use `chromPeaks` (that's what the user will use).

R/XcmsExperiment-functions.R
```diff
@@ -859,8 +863,9 @@
...
 
 #' Function to help extracting the *old* MChromatograms and XChromatograms
-#' from an XcmsExperiment. This is based on the code for XCMSnExp objects,
-#' but there could be room for improvement.
+#' from an XcmsExperiment.
+#'
+#' @param object `XcmsExperiment` or `XCMSnExp` object.
 #'
 #' @noRd
 .xmse_extract_chromatograms_old <- function(object, rt, mz, aggregationFun,
```
jorainer, line 871: this function was modified to improve performance of the `chromatogram,XcmsExperiment` call.

jorainer:
Note: unit tests fail on ubuntu (and the Bioconductor docker image) because of internal changes in R devel. Because of this *Rcpp*, *rlang* and related packages fail to install (and hence also *mzR*, *xcms* etc).

jorainer:
@sneumann @philouail , just a friendly reminder to eventually (if you find the time) have a look at this PR :)

sneumann:
Hi, any need to bump dependency versions if code was moved to `MetaboCoreUtils` ? Otherwise looks good to me. Yours, Steffen

jorainer:
Nope, no need for any version bump - this internal code that was refactored and moved to `MetaboCoreUtils` was never used within `xcms` (in any exported function).

sneumann (approved):
So if no changes / version bumps in DESCCRIPTION are needed, this looks good to me. Yours, Steffen

R/XcmsExperiment-functions.R
philouail: Can I ask the reason of why the `DataFrame` was initially used ? was it easier at the beginning ? I have not really end up using `DataFrame` until now so I'm a bit curious as to when someone decide to prefer this over `data.frame`

R/XcmsExperiment.R
```diff
@@ -1677,32 +1648,32 @@ setMethod(
         ## Populate with chrom peaks.
         nf <- nrow(fts)
         js <- seq_len(ncol(chrs))
-        pks_empty <- chromPeaks(object)[integer(), ]
-        pkd_empty <- chromPeakData(object)[integer(), ]
+        pks_empty <- .chromPeaks(object)[integer(), ]
+        pkd_empty <- as(.chromPeakData(object)[integer(), ], "DataFrame")
```
philouail, line 1652: What is happening here ?

philouail (approved):
All good for me, just have 2 questions in the comments ! also thanks for your comments explaining it helped me to understand the context of everything.

R/XcmsExperiment-functions.R
jorainer: Good question. There were 2 reasons: firstly, the `DataFrame` was used in e.g. `MSnbase` and other Bioconductor packages (e.g. `SummarizedExperiment`) and I wanted to be consistent with that, then, secondly, the `DataFrame` would also allow to store S4 objects in columns (which the `data.frame` would *officially* not). Not that we planned to store now S4 objects in `DataFrame`s - so it was mostly reason 1 why we used `DataFrame`.

Only (quite some time) later I figured out that subset or extract data from a `DataFrame`s is quite slower than for `data.frame`. Thus, recently, I prefer to use, whenever possible (internally) a `data.frame`.

R/XcmsExperiment.R
jorainer: I created new internal functions `.chromPeaks` and `.chromPeakData` to extract the content of the `chromPeaks` (or `chromPeakData`) slot from an xcms result object. We have two xcms result objects now, `XCMSnExp` and `XcmsExperiment`. The former stores the data in a quite clumsy way, and extracting the data is a little more tricky. The latter is the new preferred result object. There is also the `chromPeaks` and `chromPeakData` **method** available for each of the two result objects, but these methods do quite some additional checks and allow subset/filter the content. The new internal functions (are not expected to be called by the user) are simple helpers to extract the full chromPeaks and chromPeakData content without any filtering etc. 

I could also have created a method for that, but I decided against that and just added a function with an if/else inside.

Then, for chromPeakData I have also to ensure that the result is returned as a `DataFrame`, thus I'm calling in addition the `as(..., "DataFrame"`. For `XCMSnExp` `.chromPeakData` will already return a `DataFrame`, so nothing will happen, while for `XcmsExperiment` I'm storing the data as a `data.frame`, thus I need to convert first.

Hope this explained it?

Issue #711, opened 2024-01-15
why do_groupChromPeaks_density processing peaks with fixed size of the overlapping slices in mz dimension


DrRuiLi:
Retrieved the source code in `do_groupChromPeaks_density `
This function process peaks with fixed `binSize; numeric(1) defining the size of the overlapping slices in mz dimension.`
As code show, it split mz only based on mz range and binSize and then process in each slice.
```
  mass <- seq(peaks[1, "mz"], peaks[nrow(peaks), "mz"] + binSize, 
    by = binSize/2) 

  masspos <- findEqualGreaterM(peaks[, "mz"], mass)
  densFrom <- rtRange[1] - 3 * bw
  densTo <- rtRange[2] + 3 * bw
  densN <- max(512, 2 * 2^(ceiling(log2(diff(rtRange)/(bw/2)))))
  endIdx <- 0
  message("Processing ", length(mass) - 1, " mz slices ... ", 
    appendLF = FALSE)
  resL <- vector("list", (length(mass) - 2))
  for (i in seq_len(length(mass) - 2)) {
    startIdx <- masspos[i]
    endIdx <- masspos[i + 2] - 1
    if (endIdx - startIdx < 0) 
      next
    resL[[i]] <- .group_peaks_density(peaks[startIdx:endIdx, 
      , drop = FALSE], bw = bw, densFrom = densFrom, densTo = densTo, 
      densN = densN, sampleGroups = sampleGroups, sampleGroupTable = sampleGroupTable, 
      minFraction = minFraction, minSamples = minSamples, 
      maxFeatures = maxFeatures, sleep = sleep)
  }
```
But the width of mz distribution is related to absolute mz value (such as 20 ppm), not a fixed value.
The same problem exist in `do_groupChromPeaks_nearest` with param `absMz `
Why not slice mz in a mz value dependant manner

jorainer:
That's a good question. It was implemented that way in the original implementation - I would maybe also implement it in the way you suggested. For TOF instruments (that have a m/z dependent measurement error) having a fixed m/z bin size is indeed not ideal, but I also don't think it makes a huge difference. IMHO, by considering both m/z and rt it should be unlikely that signal from different compounds will be grouped together.

DrRuiLi:
[...]

PR #712, opened 2024-01-16
Add possibility to perform peak density correspondence considering m/z dependent margins

jorainer:
This PR adds the possibility to perform peak density-based correspondence using m/z-dependent bin sizes. The `PeakDensityParam` gains a parameter `ppm` with a default of `ppm = 0` in which case the original approach is used that groups chrom peaks in equal(fixed)-sized bins along the m/z dimension. By setting `ppm` to a value larger than 0 m/z-dependent bin sizes will be used instead: the bins along m/z increase with `ppm` of the m/z value hence better modeling the m/z dependent measurement precision/error of e.g. TOF instruments.

See also issue #711 for discussion and https://github.com/sneumann/xcms/issues/711#issuecomment-1893389759 for the consequence of such correspondence analysis.

### DIFF
--- a/R/AllGenerics.R
+++ b/R/AllGenerics.R
@@ -588,7 +588,7 @@ setGeneric("family<-", function(object, value) standardGeneric("family<-"))
 #'
 #' The EIC data of a feature is extracted from every sample using the same
 #' m/z - rt area. The EIC in a sample does thus not exactly represent the
-#' signal of the actually identified chromatographi peak in that sample.
+#' signal of the actually identified chromatographic peak in that sample.
 #' The [chromPeakChromatograms()] function would allow to extract the actual
 #' EIC of the chromatographic peak in a specific sample. See also examples
 #' below.
--- a/R/AllGenerics.R
+++ b/R/AllGenerics.R
@@ -1315,7 +1315,9 @@ setGeneric("group", function(object, ...) standardGeneric("group"))
 #'     sequence of values from the smallest to the larges m/z value with a
 #'     constant bin size of `binSize`. For `ppm` > 0 the size of each bin is
 #'     increased in addition by the `ppm` of the (upper) m/z boundary of the
-#'     bin.
+#'     bin. The maximal bin size (used for the largest m/z values) would then
+#'     be `binSize` plus `ppm` parts-per-million of the largest m/z value of
+#'     all peaks in the data set.
 #'
 #' @param param The parameter object selecting and configuring the algorithm.
 #'
--- a/R/AllGenerics.R
+++ b/R/AllGenerics.R
@@ -569,32 +569,29 @@ setGeneric("family<-", function(object, value) standardGeneric("family<-"))
 #' [XCMSnExp-class] object. The function returns for each feature the
 #' extracted ion chromatograms (along with all associated chromatographic
 #' peaks) in each sample. The chromatogram is extracted from the m/z - rt
-#' region including all chromatographic peaks of that features. This region is
-#' by default, with `mzmin = min`, `mzmax = max`, `rtmin = min` and
-#' `rtmax = max` defined using the **ranges** of `"mzmin"`, `"mzmax"`,
-#' `"rtmin"`, `"rtmax"` of all chromatographic peaks of the feature. For some
-#' features, and depending on the data, the m/z and rt range can thus be
-#' relatively large. The ranges could be restricted by using a different
-#' function to define them, e.g. by setting `mzmin = median` and
-#' `mzmax = median` in which case the median `"mzmin"` and `"mzmax"` values
-#' of all chromatographic peaks would be used.
+#' region that includes **all** chromatographic peaks of a feature. By default,
+#' this region is defined using the range of the chromatographic peaks' m/z
+#' and retention times (with `mzmin = min`, `mzmax = max`, `rtmin = min` and
+#' `rtmax = max`). For some features, and depending on the data, the m/z and
+#' rt range can thus be relatively large. The boundaries of the m/z - rt
+#' region can also be restricted by changing parameters `mzmin`, `mzmax`,
+#' `rtmin` and `rtmax` to a different functions, such as `median`.
 #'
 #' By default only chromatographic peaks associated with a feature are
-#' included. For `object` being a `XCMSnExp` object parameter `include`
-#' allows also to return all chromatographic peaks with their apex
-#' position within the selected region (`include = "apex_within"`) or any
-#' chromatographic peak overlapping the m/z and retention time range
-#' (`include = "any"`).
+#' included in the returned [XChromatograms] object. For `object` being an
+#' `XCMSnExp` object parameter `include` allows also to return all
+#' chromatographic peaks with their apex position within the selected
+#' region (`include = "apex_within"`) or any chromatographic peak overlapping
+#' the m/z and retention time range (`include = "any"`).
 #'
 #' @note
 #'
-#' The **same** m/z and rt boundaries are used on every sample to extract
-#' the ion chromatogram. The EIC might thus not exactly represent the actual
-#' EIC of each individual chromatographic peak (i.e. signal for the ion in one
-#' specific sample), since the m/z and rt boundaries might be slightly
-#' different across samples. The [chromPeakChromatograms()] function could be
-#' used to extract the actual EIC of the chromatographic peak in a specific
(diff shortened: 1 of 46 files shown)


R/do_groupChromPeaks-functions.R
```diff
@@ -120,9 +130,10 @@ do_groupChromPeaks_density <- function(peaks, sampleGroups,
     rtRange <- range(peaks[, "rt"])
 
     ## Define the mass slices and the index in the peaks matrix with an mz
-    ## value >= mass[i].
-    mass <- seq(peaks[1, "mz"], peaks[nrow(peaks), "mz"] + binSize,
-                by = binSize / 2)
+    ## value >= mass[i]. If ppm != 0 the size of the individual bins will
+    ## be dependend on the m/z value.
+    mass <- .breaks_ppm(peaks[1, "mz"], peaks[nrow(peaks), "mz"] + binSize,
+                       by = binSize / 2, ppm = ppm / 2)
```
jorainer, line 136: here we need to use `ppm / 2`, same as `binSize / 2` because the code below considers always two of the defined bins.

jorainer:
Commit https://github.com/sneumann/xcms/pull/712/commits/c8a27d59335b09506645a749173225b1900b4ffc in addition adapts (actually fixes) the `featureArea` function that is used to extract the m/z and rt boundaries for each feature. This function is used for `featureChromatograms`, gap-filling with `ChromPeakAreaParam` and EIC-based feature grouping. The function was now also better documented, which results in a quite large number of changes.

R/AllGenerics.R
```diff
@@ -569,19 +569,30 @@ setGeneric("family<-", function(object, value) standardGeneric("family<-"))
 #' [XCMSnExp-class] object. The function returns for each feature the
 #' extracted ion chromatograms (along with all associated chromatographic
 #' peaks) in each sample. The chromatogram is extracted from the m/z - rt
-#' region including all chromatographic peaks of that features (i.e. based on
-#' the ranges of `"mzmin"`, `"mzmax"`, `"rtmin"`, `"rtmax"` of all
-#' chromatographic peaks of the feature).
+#' region that includes **all** chromatographic peaks of a feature. By default,
```
jorainer, line 572: This is the updated documentation of the `featureChromatograms` function, explaining that the EICs are *globally* defined for a feature, i.e. the EIC is extracted for the same m/z - rt area from all samples. The EIC in a specific sample does thus not necessarily represent *exactly* the chromatographic peak identified in that sample

R/XcmsExperiment-functions.R
```diff
@@ -933,15 +934,50 @@
...
+#'
+#' @param features `character` with the IDs of the features. Mandatory!
+#'
+#' @return `matrix` with columns `"mzmin"`, `"mzmax"`, `"rtmin"`, `"rtmax"`
+#'     defining the range of
+#'
+#' @author Johannes Rainer
+#'
+#' @noRd
+.features_ms_region <- function(x, mzmin = median, mzmax = median,
```
jorainer, line 963: the new function that defines the m/z - rt area considering **all** chrom peaks of a feature.

R/XcmsExperiment.R
```diff
@@ -1623,12 +1621,13 @@ setMethod(
     "featureChromatograms", "XcmsExperiment",
     function(object, expandRt = 0, expandMz = 0, aggregationFun = "max",
              features = character(), return.type = "XChromatograms",
-             chunkSize = 2L, ..., progressbar = TRUE, BPPARAM = bpparam()) {
+             chunkSize = 2L, mzmin = min, mzmax = max, rtmin = min,
```
jorainer, line 1624: add parameters `mzmin`, `mzmax`, `rtmin` and `rtmax` to give the user the possbility to customize/change.

R/functions-Params.R
```diff
@@ -349,12 +349,13 @@ MergeNeighboringPeaksParam <- function(expandRt = 2, expandMz = 0, ppm = 10,
 }
 
 #' @rdname fillChromPeaks
-ChromPeakAreaParam <- function(mzmin = function(z) quantile(z, probs = 0.25),
-                               mzmax = function(z) quantile(z, probs = 0.75),
-                               rtmin = function(z) quantile(z, probs = 0.25),
-                               rtmax = function(z) quantile(z, probs = 0.75)) {
-    new("ChromPeakAreaParam", mzmin = mzmin, mzmax = mzmax, rtmin = rtmin,
-        rtmax = rtmax)
+ChromPeakAreaParam <-
```
jorainer, line 352: using `names = FALSE` skips setting the names on the returned value - can have a (little) impact on performance.

R/functions-XCMSnExp.R
```diff
@@ -2014,47 +2014,6 @@ setMethod("hasFilledChromPeaks", "XCMSnExp", function(object) {
...
-#'     defining the range of
-#'
-#' @author Johannes Rainer
-#'
-#' @noRd
-.features_ms_region <- function(x, mzmin = median, mzmax = median,
-                                rtmin = median, rtmax = median,
-                                msLevel = unique(msLevel(x)),
-                                features = character()) {
-    pk_idx <- featureValues(x, value = "index", method = "maxint",
```
jorainer, line 2036: with this code, for each feature, only the chrom peak with the highest intensity was used to define the m/z and rt area.

R/methods-Params.R
```diff
@@ -1002,6 +1002,13 @@ setReplaceMethod("maxFeatures", "PeakDensityParam", function(object, value) {
         return(object)
 })
 
+#' @rdname groupChromPeaks
+setMethod("ppm", "PeakDensityParam", function(object) {
+    if (.hasSlot(object, "ppm"))
```
jorainer, line 1007: backward compatibility - so, even loading an old (saved) instance of `PeakDensityParam` supports the `ppm` method - even if the slot is not present.

R/methods-group-features.R
```diff
@@ -697,6 +697,12 @@ plotFeatureGroups <- function(x, xlim = numeric(), ylim = numeric(),
 #'
 #' @note
 #'
+#' At present the [featureChromatograms()] function is used to extract the
```
jorainer, line 700: just explaining that the EICs are not *exactly* representing the chrom peaks.

src/fastMatch.c
```diff
@@ -29,7 +29,7 @@ SEXP fastMatch(SEXP x, SEXP y, SEXP xidx, SEXP yidx, SEXP xolength, SEXP tol) {
 
     struct idxStruct * pidxS =  (struct idxStruct *) calloc(nx,  sizeof(struct idxStruct));
     if (pidxS == NULL)
-        error("fastMatch/calloc: memory could not be allocated ! (%llu bytes)\n", nx  * sizeof(struct idxStruct) );
+        error("fastMatch/calloc: memory could not be allocated ! (%lu bytes)\n", nx  * sizeof(struct idxStruct) );
```
jorainer, line 32: that's to fix the compiler warnings during `R CMD check`.

R/AllGenerics.R
```diff
@@ -1235,10 +1284,16 @@ setGeneric("group", function(object, ...) standardGeneric("group"))
...
 #'   (Smith 2006) that groups chromatographic peaks along the retention time
-#'   axis within slices of (partially overlapping) m/z ranges. All peaks (from
-#'   the same or from different samples) with their apex position being close
-#'   on the retention time axis are grouped into a LC-MS feature. See in
-#'   addition [do_groupChromPeaks_density()] for the core API function.
+#'   axis within slices of (partially overlapping) m/z ranges. By default,
+#'   these m/z ranges (bins) have a constant size. By setting `ppm` to a value
+#'   larger than 0, m/z dependent bin sizes can be used instead (better
+#'   representing the m/z dependent measurement error of some MS instruments).
+#'   Setting `ppm` to a value different than 0 results thus in m/z dependent
```
philouail: I think this sentence repeats the previous one ?

philouail (approved):
This looks great ! Can't wait to test it.

R/AllGenerics.R
jorainer: Ah, yes, thanks! I will cut/reformulate.

sneumann (approved):
Hi, great to see the new code, thorough evaluation in the issue and updated docs and tests. Yours, Steffen

jorainer:
thanks for merging @sneumann !

PR #713, opened 2024-01-24
Addition of method to filter features based on quality assessment

philouail:
Here is the implementation of the `filterFeatures` function with different quality assessment methods. I also wrote the method so that they can handle both `XcmsResult` objects and `SummarisedExperiment` objects. 

@jorainer I have updated the `PercentMissingFilter` so that now it filters based on the different sample types using `factor`. 

I tried to make the documentation comprehensive but don't hesitate to tell me if something is not clear or not explained in enough details for the users. 

These different classes/methods are based on the quality assessment function from  MetaboCoreUtils. So some information can also be found in this package. 

Thanks in advance for the feedback !


DESCRIPTION
```diff
@@ -62,7 +62,8 @@ Imports:
     Spectra (>= 1.13.2),
     progress,
     multtest,
-    jsonlite
+    jsonlite,
+    MetaboCoreUtils
```
jorainer: can you please add `(>= 1.11.2)` next to `MetaboCoreUtils`? just to ensure that the minimal required version is available (the one that provides the functions you import).

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
+#' @title Filtering of features based on conventional quality assessment
+#'
+#' @description
+#'
+#' When dealing with metabolomics results, it is often necessary to filter
+#' features based on certain criteria. These criteria are typically derived
+#' from statistical formulas applied to full rows of data, where each row
+#' represents a feature.
+#' The `filterFeatures` function filters features based on these conventional
+#' quality assessment criteria. Multiple type of filtering are implemented and\
```
jorainer: typos: `type` -> `types` and remove the `\`.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' Supported `filter` arguments are:
+#'
+#' - [`RsdFilter`]: Calculates the relative standard deviation
+#'  (i.e. coefficient of variation) of each features in QC (Quality Control)
+#'  samples and filter them in the input object according to a provided
+#'  threshold.
+#'
+#' - [`DratioFilter`]: Computes the D-ratio or *dispersion ratio*, defined as
+#'  the standard deviation for QC samples divided by the standard deviation for
+#'  biological test samples, for each features and filter them according to a
```
jorainer: `features` -> `feature`

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#'  provided threshold
+#'
+#' - [`PercentMissingFilter`]: Determine the percentage of missing values for
+#'  each features in  and filters them according to a provided threshold.
```
jorainer: ... `each feature in the various sample groups and ...`

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#'
+#' - [`BlankFlag`]: Identifies features where the mean of test samples
```
jorainer: `... the mean abundance in test samples ...`

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' filtering. It can be one of the following classes: [`RsdFilter`],
+#' [`DratioFilter`], [`PercentMissingFilter`] or [`BlankFlag`].
+#'
+#' @param assay For filtering of `SummarizedExperiment` objects only. Indicates
+#' which assay the filtering will be based on. Note that the features for the
+#' entire object will be removed, but the computations are performed on a single
+#' assay. Default is 1, which mean the first assay of the `object` will
+#' be evaluated.
+#'
+#' @param ... Optional parameters.
```
jorainer: `... Optional parameters. For `object` being an `XcmsExperiment`: parameters for the [featureValues()] call.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' PMCID: PMC5960010.
+#'
+#' @examples
+#' ## See the vignettes for more detailed examples
+#' library(MsExperiment)
+#'
+#' ## Load a test data set with features
+#' test_xcms <- loadXcmsData()
+#'
+#' ## Set up parameter to filter based on coefficient of variation
```
jorainer: would maybe also be nice to specify the threshold and then say which features would be removed by the filter.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' rsd_filter <- RsdFilter(qcIndex = sampleData(test_xcms)$sample_type == "QC")
+#'
+#' filtered_data_rsd <- filterFeatures(object = test_xcms, filter = rsd_filter)
+#'
+#' ## Set up parameter to filter based on D-ratio and filter
+#' dratio_filter <- DratioFilter(qcIndex = sampleData(test_xcms)$sample_type == "QC",
```
jorainer: Also here, maybe good to define a threshold and describe which features will be removed by the call.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#'                               studyIndex = sampleData(test_xcms)$sample_type == "study")
+#'
+#' filtered_data_dratio <- filterFeatures(object = test_xcms,
+#' filter = dratio_filter)
+#'
+#' ## Set up parameter to filter based on the percent of missing data
```
jorainer: Maybe also say *Parameter `f` should represent the sample group of samples.` and then also set the threshold and say which features would be kept (or removed - whichever is easier to describe).

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#'
+#' @name RsdFilter
+#'
+#' @export
+#'
+#' @family Filter features in xcms
+#'
+#' @description
+#' The `RsdFilter` class and methods enable users to filter features from an
+#' `XcmsExperiment` or `SummarizedExperiment` object based on their coefficent
```
jorainer: *... on their relative standard deviation (coefficient of variation) ...*

Also add maybe:

Note that it is assumed that the abundance values are in **natural** scale. Abundances in log scale should be first transformed to natural scale before calculating the RSD.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' of variation for a specified threshold.
+#'
+#' This `filter` is part of the possible dispatch of the generic function
+#' `filterFeatures`. Features *above* the user-input threshold will be
```
jorainer: Is it `>` or `>=`? I would maybe add in brackets which of the two is used.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' removed from the entire dataset.
+#'
+#' @param threshold `numeric` value representing the threshold. Features with a
+#' coefficient of variation higher than this will be removed from the entire
```
jorainer: Also here - is it `>` or `>=` - better to tell which is the case.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' dataset.
+#'
+#' @param qcIndex `integer` (or logical) vector corresponding to the indexes of
```
jorainer: I think it's called `indices` not `indexes`? but not sure.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+    new("RsdFilter", threshold = threshold, qcIndex = qcIndex, na.rm = na.rm,
+        mad = mad)
+}
+
+#' @rdname RsdFilter
+setMethod("filterFeatures",
+          signature(object = "XcmsResult",
+                    filter = "RsdFilter"),
+          function(object, filter){
+              if (length(filter@qcIndex) < 1 |
```
jorainer: I think this test is not correct? the values need to be within 1 to the length of object, thus I would use here:

`if (!all(filter@qcIndex %in% seq_along(object)))`

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+                  length(filter@qcIndex) > length(object))
+                  stop("'qcIndex' must be within object length range")
+            vals <- featureValues(object)[, filter@qcIndex]
```
jorainer: I would here also pass the `...`: `vals <- featureValues(object, ...)[, filter@qcIndex]`

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+        qcIndex <- as.integer(qcIndex)
+    new("RsdFilter", threshold = threshold, qcIndex = qcIndex, na.rm = na.rm,
+        mad = mad)
+}
+
+#' @rdname RsdFilter
+setMethod("filterFeatures",
+          signature(object = "XcmsResult",
+                    filter = "RsdFilter"),
+          function(object, filter){
```
jorainer: add here also the `...`: `function(object, filter, ...) {` then the user can pass parameters to the `featureValues` function.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+              if (length(filter@qcIndex) < 1 |
+                  length(filter@qcIndex) > length(object))
+                  stop("'qcIndex' must be within object length range")
+            vals <- featureValues(object)[, filter@qcIndex]
+            vals <- rowRsd(x = vals,  na.rm = filter@na.rm, mad = filter@mad)
+            fts_idx <- which(vals <= filter@threshold)
+            message(paste(length(vals) - length(fts_idx),
```
jorainer: no need to use `paste` with `message`, `stop` or `warning`, please remove the `paste(`.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+                          "features were removed"))
+            ph <- XProcessHistory(param = filter,
+                                  date. = date(),
+                                  type. = .PROCSTEP.FEATURE.FILTERING,
+                                  fileIndex = seq_along(object))
+            object <- addProcessHistory(object, ph)
+            object <- filterFeatureDefinitions(object, features = fts_idx)
```
jorainer: maybe here remove the `object <- `, so simply end with `filterFeatureDefinitions(object, features = fts_idx)`. Could well be that R is making (yet) another copy of the data with the `object <- ` statement (it copies it during  the call, then again in the assignment). will not be an issue for small data sets, but could be a problem with very large data sets.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+          }
+)
+
+#' @rdname RsdFilter
+setMethod("filterFeatures",
+          signature(object = "SummarizedExperiment",
+                    filter = "RsdFilter"),
+          function(object, filter, assay = 1){
+              if (length(filter@qcIndex) < 1 | length(filter@qcIndex) > ncol(object))
```
jorainer: Also here, I would change the condition to

`if (!all(filter@qcIndex) %in% seq_len(ncol(object)))` ... I guess here it's saver to use `ncol(object)` not `length(object)` - for `SummarizedExperiment` `nrow` is the number of features/rows and `ncol` the number of samples/columns - don't know what `length` would return. For `MsExperiment` we don't consider `rows` (features) because they are not really defined there - so `length` of an `MsExperiment` is the number of samples.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' @export
+#'
+#' @family Filter features in xcms
+#'
+#' @description
+#' The `DratioFilter` class and method enable users to filter features from an
+#' `XcmsExperiment` or `SummarizedExperiment` object based on the D-ratio or
+#' *dispersion ratio*. This is defined as the standard deviation for QC
+#' samples divided by the standard deviation for biological test samples, for
+#' each feature of the object.
```
jorainer: I would add here *... of the object (Broadhurst et al.).* and then add also the reference in this help page.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#'
+#' This `filter` is part of the possible dispatch of the
+#' generic function `filterFeatures`.  Features *above* the user-input
+#' threshold will be removed from the entire dataset.
+#'
+#' @param threshold `numeric` value representing the threshold. Features with a
+#' coefficient of variation higher than this will be removed from the entire
```
jorainer: should be *D-ratio* not *coefficient of variation*?

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' dataset.
+#'
+#' @param qcIndex `integer` (or `logical`) Vector corresponding to the indexes of
```
jorainer: *Vector* -> *vector*

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' QC samples.
+#'
+#' @param studyIndex `integer` (or `logical`) Vector corresponding of the
```
jorainer: also here lower case for *Vector*

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+        studyIndex <- as.integer(studyIndex)
+    new("DratioFilter", threshold = threshold, qcIndex = qcIndex,
+        studyIndex = studyIndex, na.rm = na.rm, mad = mad)
+}
+
+#' @rdname DratioFilter
+setMethod("filterFeatures",
+          signature(object = "XcmsResult",
+                    filter = "DratioFilter"),
+          function(object, filter){
```
jorainer: would also add here the `...`: `function(object, filter, ...) {` to allow passing parameters to the `featureValues`

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+              if (length(filter@studyIndex) < 1 |
```
jorainer: Please fix also these tests for `@studyIndex` and `@qcIndex`.

Note that, since you're testing this in every function now, it could also maybe make sense to implement a helper function? something like:

```r
.check_index_range <- function(x, l, name = "") {
    if (!all(x %in% seq_len(l)))
        stop(name, " should be between 1 and ", l)
}
```

and then call this e.g. here

```r
function(object, filter, ...) {
    .check_index_range(filter@studyIndex, length(object), name = "studyIndex")
    .check_index_range(filter@qcIndex, length(object), name = "qcIndex")
    ....

```

for the `SummarizedExperiment` you would then need to call it with `ncol(object)` instead of `length(object)`.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+                  length(filter@studyIndex) > length(object))
+                  stop("'studyIndex' must be within object length range")
+              if (length(filter@qcIndex) < 1 |
+                  length(filter@qcIndex) > length(object))
+                  stop("'qcIndex' must be within object length range")
+              x <- featureValues(object)[, filter@studyIndex]
```
jorainer: here I would pass the `...` along: `x <- featureValues(object, ...)[, filter@studyIndex]`, same then also below for `y`.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+              y <- featureValues(object)[, filter@qcIndex]
+              vals <- rowDratio(x = x, y = y,
+                                na.rm = filter@na.rm,
+                                mad = filter@mad)
+              fts_idx <- which(vals <= filter@threshold)
```
jorainer, line 350: just to be sure - we remove the features that have a D-ratio above the threshold? Or should we remove the ones with a D-ratio below?

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+              message(paste(length(vals) - length(fts_idx),
+                            "features were removed"))
+              ph <- XProcessHistory(param = filter,
+                                    date. = date(),
+                                    type. = .PROCSTEP.FEATURE.FILTERING,
+                                    fileIndex = seq_along(object))
+              object <- addProcessHistory(object, ph)
+              object <- filterFeatureDefinitions(object, features = fts_idx)
```
jorainer: also here, please remove the last `object <- ` assignment and leave just the `filterFeatureDefinition` call.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+          }
+)
+
+#' @rdname DratioFilter
+setMethod("filterFeatures",
+          signature(object = "SummarizedExperiment",
+                    filter = "DratioFilter"),
+          function(object, filter, assay = 1){
+              if (length(filter@studyIndex) < 1 |
```
jorainer: need to fix the tests here.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' @name PercentMissingFilter
+#'
+#' @export
+#'
+#' @family Filter features in xcms
+#'
+#' @description
+#' The `PercentMissingFilter` class and method enable users to filter features
+#' from an `XcmsExperiment` or `SummarizedExperiment` object based on the
+#' percentage of missing values for each features and filters them according to
```
jorainer: *missing values for each features in different sample groups and ...*

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' a provided threshold.
+#'
+#' This `filter` is part of the possible dispatch of the generic function
+#' `filterFeatures`. The features *above* the user input threshold will be
```
jorainer: Here I think we need to be more specific what exactly we remove. Isn't it something like:

*Features with a proportion of missing values higher than `threshold` in **all** sample groups will be removed (i.e. features for which the proportion of missing values is `< threshold` in at least one sample group will be retained).*

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' removed by calling the `filterFeatures` function.
+#'
+#' @param f `vector` of the same length as the `object`, specifying the sample
+#' type for each sample in the dataset. The percentage of missing values will
```
jorainer: Instead of *The percentage ...*: *The proportion of missing values per feature will be computed within each of these sample groups. Parameter `f`, if not already a `factor`, will be converted to one using the `factor` function. Samples with an `NA` as their value in `f` will be excluded from calculation.*.

feel free to reformulate if that does not make sense...

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' be computed for each of the sample types.
+#'
+#' @param threshold `numeric`  percent of accepted missing data for one feature.
```
jorainer: I think we talk about a proportion here not percentage, i.e. values between 0 and 1? then we should also explicitely state that: 

`numeric` proportion (between 0 and 1) of accepted missing values for a feature in one sample group.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' @author Philippine Louail
+#'
+#' @importFrom MetaboCoreUtils rowPercentMissing
+#'
+NULL
+
+#' @noRd
+setClass("PercentMissingFilter",
+         slots = c(threshold = "numeric",
+                   f = "character"),
```
jorainer: I would rather use here a "factor".

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+         contains = "Param",
+         prototype = prototype(
+             threshold = numeric(),
+             f = character()),
```
jorainer: `f = factor()`

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+             msg <- NULL
+             if (length(object@threshold) != 1)
+                 msg <- c("'threshold' has to be a numeric of length 1")
+             msg
+         })
+
+#' @rdname PercentMissingFilter
+#'
+#' @export
+PercentMissingFilter <- function(threshold = 30, f = character()) {
```
jorainer: also here `f = factor()`

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+    new("PercentMissingFilter", threshold = threshold, f = f)
```
jorainer, line 442: I would then check:

```r
if (!is.factor(f))
    f <- factor(f)
```

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+}
+
+#' @rdname PercentMissingFilter
+setMethod("filterFeatures",
+          signature(object = "XcmsResult",
+                    filter = "PercentMissingFilter"),
+          function(object, filter){
+              if (length(filter@f) != length(object))
+                  stop("'f' must be same lenght as object")
+              f <- factor(filter@f)
```
jorainer: I would store the parameter as a `factor` in the `filter`, thus, no need to convert here to a factor.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+              fts_idx <- c()
+              for (i in levels(f)){
+                  spl_idx <- which(f == i)
+                  vals <- rowPercentMissing(featureValues(object)[, spl_idx])
```
jorainer: again, here please use `featureValues(object, ...)`

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' @export
+PercentMissingFilter <- function(threshold = 30, f = character()) {
+    new("PercentMissingFilter", threshold = threshold, f = f)
+}
+
+#' @rdname PercentMissingFilter
+setMethod("filterFeatures",
+          signature(object = "XcmsResult",
+                    filter = "PercentMissingFilter"),
+          function(object, filter){
```
jorainer: I would add also here the `...`: `function(object, filter, ...) {`

R/method-filterFeatures.R
jorainer: hm, or is it a percent? Please clearly specify whether we are talking here about percent (values from 0 to 100) or proportion (values from 0 to 1).

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+              }
+              fts_idx <- order(unique(fts_idx))
+              message(paste(length(vals) - length(fts_idx),
+                            "features were removed"))
+              ph <- XProcessHistory(param = filter,
+                                    date. = date(),
+                                    type. = .PROCSTEP.FEATURE.FILTERING,
+                                    fileIndex = seq_along(object))
+              object <- addProcessHistory(object, ph)
+              object <- filterFeatureDefinitions(object, features = fts_idx)
```
jorainer: also here, please remove the last `object <- ` assignment.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' means that an extra column will be created in `featureDefinitions` or
+#' `rowData` called `possible_contaminants` with a logical value for each
+#' feature.
+#'
+#' @param threshold `numeric` indicates the minimum difference
+#' required between the mean of a feature in samples compared to the mean of
+#' the same feature in blanks for it to not be considered a possible
+#' contaminant. For example, the default threshold of 2 signifies that the mean
+#' of the features in samples has to be at least twice the mean in blanks for
+#' it not to be flagged as a possible contaminant.
```
jorainer: *... in blanks for it to not be flagged...*

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+        qcIndex <- as.integer(qcIndex)
+    new("BlankFlag", threshold = threshold, blankIndex = blankIndex,
+        qcIndex = qcIndex, na.rm = na.rm)
+}
+
+#' @rdname BlankFlag
+setMethod("filterFeatures",
+          signature(object = "XcmsResult",
+                    filter = "BlankFlag"),
+          function(object, filter){
```
jorainer: please add the `...` also here.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+              if (length(filter@blankIndex) < 1 |
```
jorainer: fix also the checks for `@blankIndex` and `@qcIndex`.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+                  length(filter@blankIndex) > length(object))
+                  stop("'blankIndex' must be within object length range")
+              if (length(filter@qcIndex) < 1 |
+                  length(filter@qcIndex) > length(object))
+                  stop("'qcIndex' must be within object length range")
+              x <- featureValues(object)[, filter@qcIndex]
```
jorainer: `featureValues(object, ...)`

jorainer: also below.

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+              object
+          }
+)
+
+#' @rdname BlankFlag
+setMethod("filterFeatures",
+          signature(object = "SummarizedExperiment",
+                    filter = "BlankFlag"),
+          function(object, filter, assay = 1){
+              if (length(filter@blankIndex) < 1 |
```
jorainer: also fix the checks here.

tests/testthat/test_function-filtering.R
```diff
@@ -0,0 +1,112 @@
...
+
+    #test in SummarizedExperiment object
+    res <- quantify(xmse_full)
+    filter <- RsdFilter(qcIndex = res$sample_type == "QC")
+    filtered_res <- filterFeatures(res, filter)
+    expect_lte(nrow(filtered_res), nrow(res))
+
+    #test same amount of features in sumexp and xcmsobjectg
+    expect_equal(nrow(featureDefinitions(filtered_xmse)),
+               nrow(filtered_res)) # this is not equal, normal ??
```
jorainer: hm, no, this has to be the same, if the same filter is applied.

jorainer: but it's working now, no? otherwise the unit tests would fail. so maybe remove the comment?

tests/testthat/test_function-filtering.R
```diff
@@ -0,0 +1,112 @@
...
+    filtered_xmse <- filterFeatures(xmse_full, filter)
+    expect_lte(nrow(featureDefinitions(filtered_xmse)),
+               nrow(featureDefinitions(xmse_full)))
+
+    #test error message when index too small of too large
+    filter_1 <- RsdFilter(qcIndex = integer())
+    filter_2 <- RsdFilter(qcIndex = c(sampleData(xmse_full)$sample_type == "QC", TRUE))
+    expect_error(filterFeatures(xmse_full, filter_1))
+    expect_error(filterFeatures(xmse_full, filter_2))
+
```
jorainer, line 26: maybe also add this test:

```r
filter_3 <- RsdFilter(qcIndex = c(2, 15))
expect_error(filterFeatures(xmse_full, filter_3))
```

vignettes/xcms.Rmd
```diff
@@ -1106,6 +1106,110 @@ properties of the mice analyzed (sex, age, litter mates etc).
...
+rows of data, where each row represents a feature. The `filterFeatures`
+function provides a robust solution for filtering features based on these
+conventional quality assessment criteria. It supports multiple types of
+filtering, allowing users to tailor the filtering process to their specific
+needs, all controlled by the `filter` argument. This function and its
+implementations are applicable to both `XcmsExperiment` results objects and
+`SummarizedExperiment` objects.
+
+We will demonstrate how to use the `filterFeatures` function to perform quality
+assessment and filtering on both the faahko and res objects. The `filter`
```
jorainer: *... on both the `faahko` and `res` variables defined above. The ...*

vignettes/xcms.Rmd
```diff
@@ -1106,6 +1106,110 @@ properties of the mice analyzed (sex, age, litter mates etc).
...
+argument can accommodate various types of input, each determining the
+specific type of quality assessment and filtering to be performed.
+
+The `RsdFilter` enable users to filter features based on their coefficient of
```
jorainer: *... their relative standard deviation (coefficient of variation) for ...*

vignettes/xcms.Rmd
```diff
@@ -1106,6 +1106,110 @@ properties of the mice analyzed (sex, age, litter mates etc).
...
+                        qcIndex = sampleData(faahko)$sample_type == "QC")
+
+# Apply the filter to faakho object
+filtered_faahko <- filterFeatures(object = faahko, filter = rsd_filter)
+
+# Now apply the same strategy to the res object
+rsd_filter <- RsdFilter(threshold = 0.3, qcIndex = res$sample_type == "QC")
+filtered_res <- filterFeatures(object = res, filter = rsd_filter, assay = "raw")
+```
+
```
jorainer, line 1190: Maybe say what happened by the call above.

*All features with an RSD (CV) larger than 0.3 in QC samples were thus removed from the data set.*

vignettes/xcms.Rmd
```diff
@@ -1106,6 +1106,110 @@ properties of the mice analyzed (sex, age, litter mates etc).
...
+filtered_res <- filterFeatures(object = filtered_res,
+                               filter = dratio_filter)
+```
+
+The `PercentMissingFilter` allows to filter features based on the percentage of
+missing values for each feature. This function takes as an input the parameter
+`f` which is supposed to be a vector of the length of the object with the
+sample type for each samples. The function then compute the percentage of
+missing values per sample groups and filters features based on this. Another
```
jorainer: Maybe also specify which features get removed:

*based on this. Features with a percent of missing values larger than `threshold` in **all** sample groups will be removed. ...*

vignettes/xcms.Rmd
```diff
@@ -1106,6 +1106,110 @@ properties of the mice analyzed (sex, age, litter mates etc).
...
+option, it so base this quality assessment and filtering only on QC sample.
+
+Both examples are shown below: 
+
+```{r}
+# Set up parameter `f` to filter per sample type
```
jorainer: *...per sample type excluding QC samples*

jorainer (changes requested):
Really great PR adding relevant functionality @philouail ! I have some change requests, but only minor ones.

R/method-filterFeatures.R
philouail: Yes we remove the features that are above (From the paper :"it is recommended that the acceptance criterion for D-ratio is set to, at most, < 50% (preferably much lower)".  I'm rewriting the description to make it very clear how it filters and what is gonna be removed.

R/method-filterFeatures.R
philouail: I have a question regarding that for the BlankFlag function, see the comment at the bottom :)

R/method-filterFeatures.R
philouail: Yes it's percent, i'll make it more clear !

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+              message(paste(length(featureValues(object)) - sum(vals),
+                            "features were flagged"))
+              featureDefinitions(object)$possible_contaminants <- vals
+              ph <- XProcessHistory(param = filter,
+                                    date. = date(),
+                                    type. = .PROCSTEP.FEATURE.FILTERING,
+                                    fileIndex = seq_along(object))
+              object <- addProcessHistory(object, ph)
+              validObject(object)
+              object
```
philouail, line 571: here do I need to call the object at the end, or validObject() will return the object anyway ? I think it won't work but just checking

R/method-filterFeatures.R
```diff
@@ -0,0 +1,597 @@
...
+#' ## Set up parameter to filter based on the percent of missing data
+#' missing_data_filter <- PercentMissingFilter(f = sampleData(test_xcms)$sample_type)
+#'
+#' filtered_data_missing <- filterFeatures(object = test_xcms,
+#' filter = missing_data_filter)
+#'
+#' ## Set up parameter to flag possible contaminant based on blank samples
+#' filter <- BlankFlag(
+#'     qcIndex = sampleData(test_xcms)$sample_type == "QC",
+#'     blankIndex = sampleData(test_xcms)$sample_type == "study")
```
philouail: here i'm not sure what to do, I think it might be confusing for people that i use study samples as blank samples(because the object does not have blanks). I can explain it but is it a good idea ? 
Otherwise should I had like 2 randomly generated blank sample to the object and use that as an example ?

R/method-filterFeatures.R
philouail: that's good, and probably can reuse it in other functions later.

R/method-filterFeatures.R
philouail: This does not work if filter@qcIndex == 0 . So i added another check in the function

DESCRIPTION
```diff
@@ -62,7 +62,8 @@ Imports:
     Spectra (>= 1.13.2),
     progress,
     multtest,
-    jsonlite
+    jsonlite,
+    MetaboCoreUtils (>= 1.11.2)
```
jorainer, line 66: can you please change the version for `ProtGenerics` above also to 1.35.2? I forgot that in my own PR...

vignettes/xcms.Rmd
```diff
@@ -1153,6 +1153,121 @@ properties of the mice analyzed (sex, age, litter mates etc).
...
+filtered_res <- filterFeatures(object = filtered_res,
+                               filter = dratio_filter)
+```
+
+All features with an D-ratio strictly larger than 0.5 were thus removed from
+the data set.
+
+The `PercentMissingFilter` allows to filter features based on the percentage of
+missing values for each feature. This function takes as an input the parameter
+`f` which is supposed to be a vector of the length of the object with the
```
jorainer: I think there is a *length* missing in the sentence - so it should read somehow like below - and in the subsequent sentences there are also some smaller mistakes and typos:

... which is supposed to be a vector of length equal to the length of the object (i.e. number of samples) with the sample type for each. The function then compute**s** ... ... Another option is to base...

R/method-filterFeatures.R
```diff
@@ -0,0 +1,622 @@
...
+#'
+#' - [`RsdFilter`]: Calculates the relative standard deviation
+#'  (i.e. coefficient of variation) in abundance for each feature in QC
+#'  (Quality Control) samples and filters them in the input object according to
+#'  a provided threshold.
+#'
+#' - [`DratioFilter`]: Computes the D-ratio or *dispersion ratio*, defined as
+#'  the standard deviation in abundance for QC samples divided by the standard
+#'  deviation for biological test samples, for each feature and filters them
+#'  according to a provided threshold
```
jorainer: sorry, now I'm really becoming picky - but could you please add a . at the end of the sentence?

R/method-filterFeatures.R
jorainer: great! thanks for clarifying!

R/method-filterFeatures.R
```diff
@@ -0,0 +1,622 @@
...
+              message(sum(vals, na.rm = TRUE) - length(object),
+                      " features were flagged")
+              rowData(object)$possible_contaminants <- vals
+              object
+          }
+)
+
+#' @noRd
+.check_index_range <- function(x, l, name = "") {
+    if (!all(x %in% seq_len(l)) | length(x) == 0)
```
jorainer: as discussed, I would reverse the order of the tests and use `||`, i.e. `if (length(x) == 0 || !all(x %in% seq_len(l)))`

jorainer (approved):
Great contribution @philouail ! I have only very minor things/typos left. After that I would be OK for merging - given @sneumann agrees.

sneumann (approved):
Hi, thanks a lot, nothing to add ! Yours, Steffen

Issue #716, opened 2024-01-30
Trouble extracting chromatogram plots after manualChromPeaks()


pablovgd:
Hi all, me again :) 

I just found the `manualChromPeaks` function in xcms, that _allows to manually define chromatographic peaks, integrate the intensities within the specified peak area and add them to the object's chromPeaks matrix._ 

On first sight, everything seems to work fine:

```
> res <- manualChromPeaks(data_batch, pks)
> res
Object of class XcmsExperiment 
 Spectra: MS1 (22993) 
 Experiment data: 20 sample(s)
 Sample data links:
  - spectra: 20 sample(s) to 22993 element(s).
 xcms results:
  - chromatographic peaks: 4444 in MS level(s): 1 

> head(chromPeaks(res))
              mz     mzmin     mzmax        rt rtmin rtmax       into sample        maxo
CP0001 103.07533 103.07484 103.07588 510.20761 502.2 520.2    1263941      1    317976.2
CP0002 133.04950 133.04887 133.05021 253.54175 244.8 262.8    2862221      1    425176.2
CP0003 139.05017 139.04950 139.05090  86.32060  70.2  88.2   28472714      1  19922294.0
CP0004  90.05496  90.05451  90.05541  52.51372  39.0  57.0  463882888      1 144385264.0
CP0005 118.08625 118.08567 118.08685  85.34979  72.0  90.0  998694923      1 317063936.0
CP0006 132.10189 132.10125 132.10257 129.80821 119.4 137.4 5575732931      1 568556352.0
```

However, when I want to inspect the integration of the peaks in plots of the chromatograms, I run into issues.

When I extract the chromatograms using `chromPeakChromatograms`, the resulting object displays the following error:

```
> chroms <- chromPeakChromatograms(res)
> chroms
XChromatograms with 4444 rows and 1 column
Error in FUN(X[[i]], ...) : 
  invalid class “XChromatogram” object: chromPeaks matrix does not have all required columns: rt,rtmin,rtmax,into,maxo,sn


```

Merging neighboring peaks and performing peak grouping first also leads to the same error:

```

 >      mpp <- MergeNeighboringPeaksParam(expandRt = 4)
>       res_pp <- refineChromPeaks(res, mpp)

>   pdp <- PeakDensityParam(sampleGroups = sampleData(res)$sample_type,
+                               minFraction = 0.4, bw = 30)
>       res <- groupChromPeaks(res_pp, param = pdp)
[...]

PR #719, opened 2024-02-08
fix issue #716 in XcmsExperiment-functions.R

pablovgd:
As mentioned in issue #716: I tried to implement a fix that makes sure the `sn` column also appears when using the `manualChromPeaks` function so that other function to extract and plot the `XChromatograms` objects work. Noise estimation happens using the `estimateChromNoise` function, feel free to suggest better ways to implement this.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -35,7 +35,7 @@ Authors@R: c(
 	   person(given = "Philippine", family = "Louail",
 	   email = "philippine.louail@eurac.edu",
 	   role = "aut",
-	   comment = c(ORCID = "0009-0007-5429-6846"))
+	   comment = c(ORCID = "0009-0007-5429-6846")),
 	   person(given = "Pablo", family = "Vangeenderhuysen",
 	   email = "pablo.vangeenderhuysen@ugent.be",
 	   role = "ctb",
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: xcms
-Version: 4.1.7
+Version: 4.1.8
 Title: LC-MS and GC-MS Data Analysis
 Description: Framework for processing and visualization of chromatographically
     separated and single-spectra mass spectral data. Imports from AIA/ANDI NetCDF,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -36,7 +36,6 @@ Authors@R: c(
 	   email = "philippine.louail@eurac.edu",
 	   role = "aut",
 	   comment = c(ORCID = "0009-0007-5429-6846"))
-	   )
 	   person(given = "Pablo", family = "Vangeenderhuysen",
 	   email = "pablo.vangeenderhuysen@ugent.be",
 	   role = "ctb",
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -37,6 +37,11 @@ Authors@R: c(
 	   role = "aut",
 	   comment = c(ORCID = "0009-0007-5429-6846"))
 	   )
+	   person(given = "Pablo", family = "Vangeenderhuysen",
+	   email = "pablo.vangeenderhuysen@ugent.be",
+	   role = "ctb",
+	   comment = c(ORCID = "0000-0002-5492-6904"))
+	   )
 Depends:
     R (>= 4.0.0),
     BiocParallel (>= 1.8.0),
--- a/R/XcmsExperiment-functions.R
+++ b/R/XcmsExperiment-functions.R
@@ -503,7 +503,6 @@
                                            mzCenterFun = "weighted.mean",
                                            sampleIndex = integer(),
                                            cn = character(), ...) {
-    cn = c(cn,"sn") 
     res <- matrix(NA_real_, ncol = length(cn), nrow = nrow(peakArea))
     rownames(res) <- rownames(peakArea)
     colnames(res) <- cn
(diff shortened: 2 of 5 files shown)


jorainer:
Thanks @pablovgd ! I'll have a look into this. it might be safer to introduce/require the column further upstream, but I'll check that.

R/XcmsExperiment-functions.R
```diff
@@ -525,10 +526,11 @@
                 maxi <- which.max(mat[, 2L])[1L]
                 mmz <- do.call(mzCenterFun, list(mat[, 1L], mat[, 2L]))
                 if (is.na(mmz)) mmz <- mat[maxi, 1L]
-                res[i, c("rt", "mz", "maxo", "into")] <- c(
+                res[i, c("rt", "mz", "maxo", "into","sn")] <- c(
                     rts[maxi], mmz, mat[maxi, 2L],
                     sum(mat[, 2L], na.rm = TRUE) *
-                    ((rtr[2L] - rtr[1L]) / max(1L, (length(keep) - 1L)))
+                    ((rtr[2L] - rtr[1L]) / max(1L, (length(keep) - 1L))),
+                    mat[maxi, 2L]/ estimateChromNoise(mat[, 2L])
```
jorainer: I would not calculate the signal-to-noise ratio here. The reason is that this function will be mostly used to do the actual peak integration (e.g. also for gap-filling), thus, the `mat` will contain only signal of the actual chrom peaks and none, or very few data points, outside of that. the `estimateChromNoise` would therefore overestimate the noise and the `"sn"` value would be wrong.

So here, I think it would be better if we report `NA` for the `"sn"` - and since we're initializing the matrix above with `NA` we better don't insert anything for the `"sn"` column.

R/XcmsExperiment-functions.R
```diff
@@ -503,6 +503,7 @@
                                            mzCenterFun = "weighted.mean",
                                            sampleIndex = integer(),
                                            cn = character(), ...) {
+    cn = c(cn,"sn")
```
jorainer: will be possibly problematic to add `"sn"` here - if `chromPeaks` were detected previously then there should already be a `"sn"` in `cn` - and here we would add an additional one. Better to ensure upstream of this function that `cn` contains `"sn"`. For objects, after calling `findChromPeaks`, that column will be present and hence passed with `cn`.

jorainer (changes requested):
Thanks @pablovgd for looking through the whole code and nailing the problem down. 

I think however that it would be better to fix the problem upstream - the `"sn"` column should be always reported/present in the `chromPeaks` matrix. In fact, after calling `findChromPeaks` there will be a *real* chrom peaks matrix with that column. Only, for an empty `XcmsExperiment` object `chromPeaks` will return a chrom peak matrix that lacks this column (it returns an 0-row matrix with all default column names, except `"sn"`). I think it's better if we fix the issue there.

This means: please ensure that the `.empty_chrom_peaks` function https://github.com/sneumann/xcms/blob/devel/R/XcmsExperiment.R#L704-L710 returns also a `"sn"` column.

That *should* actually fix your issue and you can restore the changes you made to the `.chrom_peak_intensity_centWave` function.

Also, please increase the version in `DESCRIPTION` to 4.1.8 and also add an entry to the inst/NEWS file ideally linking also to your original issue.

Also, would be really great if you could in addition add a unit test for this fix. Maybe exactly for your use case, maybe adding to `"manualChromPeaks,XcmsExperiment"` in tests/testthat/text_XcmsExperiment.R: you could call `manualChromPeaks` on the `mse` object and then ensure that the `chromPeaks` has also the `"sn"` column or something along these lines.

Thanks again for looking into this!

jorainer:
Ah, and if that's your first pull request - just make changes in your fork and push them (to your fork), this will automatically update this PR. And click on "Re-request Review" on the right in the "Reviewers" panel to re-request my review.

pablovgd:
Thanks for the feedback Johannes! I'll take a look to make the changes you requested and will re-request review when ready.

pablovgd:
Alright, I think I managed to implement all changes you requested. Thanks again for the feedback.

jorainer (approved):
Looks good to me! Thanks for the contribution!

I leave it to @sneumann to merge and then push also to Bioconductor.

Thanks!

jorainer:
Ah @pablovgd , and one last change request: please add yourself as a contributor ("ctb") to the DESCRIPTION file. The bug fix along with the unit tests *qualifies* you as a contributor following the [RforMS contribution guidelines](https://rformassspectrometry.github.io/RforMassSpectrometry/articles/RforMassSpectrometry.html#contributions).

pablovgd:
@jorainer  Ah, great, done!

DESCRIPTION
```diff
@@ -36,6 +36,10 @@ Authors@R: c(
 	   email = "philippine.louail@eurac.edu",
 	   role = "aut",
 	   comment = c(ORCID = "0009-0007-5429-6846"))
```
jorainer: @pablovgd , please add a `,` here at the end of the line.

pablovgd: My bad, done!

sneumann:
Thanks everyone, merging now and pushing to BioC. Yours, Steffen

sneumann:
Hi,
before pushing to BioC I ran a local `R CMD check:  and got:
```
  Running ‘testthat.R’
 ERROR
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
   - maxPeakwidth: [1] 13.2
  Object of class:  MergeNeighboringPeaksParam 
...
  Error in parse(con, n = -1, srcfile = srcfile, encoding = "UTF-8") : 
    test_XcmsExperiment.R:1001:5: unexpected symbol
  1001:     res2
            ^
  Calls: test_check ... doTryCatch -> lapply -> FUN -> source_file -> parse
  Execution halted
```
and there seems to be a missing closing bracket in 
https://github.com/sneumann/xcms/blob/b1ae81942dad37169c21a6c717d4ce7480b64c07/tests/testthat/test_XcmsExperiment.R#L999
So we should also check the output of the GitHub actions:
https://github.com/sneumann/xcms/actions/runs/7957720356/job/21721103315#step:21:652

I fixed the missing bracket, working now, and pushed to BioC.
Yours, Steffen

jorainer:
Ah, oh, I'll also check and eventually fix.

jorainer:
Ah, didn't see that you already fixed. can only confirm that it works now.

PR #722, opened 2024-02-27
Ensure backward compatibility for Parameter classes that gained new slots

jorainer:
Ensure backward compatibility for Parameter classes that gained new slots: `as.list` first updates the object to the newest definition avoiding thus an error if the parameter object was e.g. loaded from and RData object.


jorainer:
Updated with restructuring of NAMESPACE and new save of bundled objects (issue #729)

jorainer:
@sneumann , maybe good to merge this PR soonish - since it fixes several things/NAMESPACE etc

PR #732, opened 2024-03-14
`LamaParama` implementation for `AdjustRtime()`: correcting alignment based on external dataset.

philouail:
Here I propose the implementation of a new parameter for `AdjustRtime()` called `LamaParama` which compute adjustment of retention file by file based on landmarks features (`lamas`) of an external dataset. 

I also implemented some diagnostic function `matchLamaChromPeaks()`, `summarizeLamaMatch` to evaluate matching of 
chromatographic peaks against `lamas` before adjustment. 

Of note, I also wrote a new interpolation function `.force_sorted()` which could be reused in other cases and might be moved to another package. 

I also wrote an example in the vignette describing everything. 

Welcoming any feedback. I'm also not sure that the functions/documentations are in the right files, would appreciate input for that. For example the diagnostic functions should probably have their own documentation ? I'm not completely sure.


R/AllGenerics.R
```diff
@@ -60,6 +60,35 @@ setGeneric("addProcessHistory", function(object, ...)
 #'   in `param`. See also [do_adjustRtime_peakGroups()] for the core API
 #'   function.
 #'
+#' - `LamaParama`: This function performs retention time correction by aligning
+#'   chromatographic data to an external reference dataset (method by Carl
```
jorainer: Maybe write here instead of *method by Carl Brunius* -> *concept and initial implementation by Carl Brunius*. Also, there is a white space missing between this and the next sentence.

R/DataClasses.R
```diff
@@ -1467,6 +1467,47 @@ setClass("PeakGroupsParam",
              else TRUE
          })
 
+#' @slot rtMap,nChrompeaks slots that should not be accessed by the user.
```
jorainer: I would actually not document the class itself, but only add `LamaParama-class` to the `@aliases` of the *main* Lama documentation.

R/AllGenerics.R
```diff
@@ -60,6 +60,35 @@ setGeneric("addProcessHistory", function(object, ...)
...
+#'   Brunius).The process involves identifying and aligning peaks within the
+#'   experimental chromatographic data, represented as an `XcmsExperiment`
+#'   object, to a predefined set of landmark features called "lamas". These
+#'   landmark features are characterized by their mass-to-charge ratio (m/z)
+#'   and retention time.
+#'
+#'   The alignment algorithm matches chromatographic peaks from the experimental
```
jorainer: I would suggest to create a dedicated, separate, documentation for the *topic* "alignment against external reference", where you provide the details of this method and where also all other related functions and methods should be described. Here you could then maybe say: *see [LamaParama()] for more information on the method*.

I would then maybe put that other documentation maybe best into *do_adjustRtime_functions.R*?

The title could be something like "Landmark-based alignment: aligning a dataset against an external reference"? I would then describe there the method (like below) and also the other functions.  You could have the functions simply listed and briefly described like:

```r
Alignment can be performed using the `adjustRtime()` methods with ...

Other related functionality to this method:

- `LamaParama()`: create the parameter object for the ...

- `matchLamasChromPeaks()`: ...

...

```

I would maybe just add this documentation there, specifying e.g. `#' @name LamaParama` and ending with `NULL` at the end. For all other functions/methods you want to document there you put then `#' @rdname LamaParama` before the function, but I would keep all the documentation, description etc in this other *main* documentation.

Also, you would need to then document all parameters to all functions listed there - ideally alphabetically ordered?

R/do_adjustRtime-functions.R
```diff
@@ -582,3 +582,283 @@ adjustRtimeSubset <- function(rtraw, rtadj, subset,
...
+#'
+#' @author Philippine Louail, Carl Brunius
+#'
+#' @rdname adjustRtime
+matchLamasChromPeaks <- function(object, param, BPPARAM = bpparam()){
+    if (!hasChromPeaks(object))
+        stop("'object' needs to have detected ChromPeaks. ",
+             "Run 'findChromPeaks()' first.")
+    cp_raw <- split.data.frame(chromPeaks(object)[, c("mz", "rt")],
+                               chromPeaks(object)[, "sample"])
```
jorainer: I would suggest here the following instead of `chromPeaks(object)[, "sample"]`:

```r
as.factor(chromPeaks(object)[, "sample"], levels = seq_along(object))
```

the reason is that it *could* be that there is a sample without detected chrom peaks, thus, just `split.data.frame` based on the `"sample"` column will return a `list` of length that is less than the number of samples you have. By specifying also the `levels` to the factor, you force the split to also return something (I guess it's either a 0-row data frame or NULL) for that sample. So, the length of the returned list matches the number of samples and the following code (e.g. the `mapply` later doesn't break) with a cryptic error message.

R/XcmsExperiment.R
```diff
@@ -1357,6 +1357,68 @@ setMethod(
...
+                 "Run 'findChromPeaks()' first")
+        if (hasAdjustedRtime(object))
+            stop("Alignment results already present. Please either remove ",
+                 "them with 'dropAdjustedRtime' in order to perform an ",
+                 "alternative, new, alignment, or use 'applyAdjustedRtime'",
+                 " prior 'adjustRtime' to perform a second round of ",
+                 "alignment.")
+        fidx <- as.factor(fromFile(object))
+        rt_raw <- split(rtime(object), fidx)
+        cp_raw  <- split.data.frame(chromPeaks(object)[, c("mz", "rt")],
```
jorainer: you can remove this - we don't need the `cp_raw` anymore in this function.

R/XcmsExperiment.R
```diff
@@ -1357,6 +1357,68 @@ setMethod(
         object
 })
 
+#'@rdname adjustRtime
+setMethod(
+    "adjustRtime", signature(object = "XcmsExperiment", param = "LamaParama"),
+    function(object, param, BPPARAM = bpparam(), ...) {
+        if(!hasChromPeaks(object))
```
jorainer: please add a white space between `if` and `(` - here and also below - and in other functions too :)

R/do_adjustRtime-functions.R
```diff
@@ -582,3 +582,283 @@ adjustRtimeSubset <- function(rtraw, rtadj, subset,
...
+#'
+#' Compute a model representing the relationship between observed and reference
+#' retention times (`rt_map`) and adjust the raw retention times (`rt_raw`)
+#' based on this.
+#'
+#' @param rt_map `data.frame` with *reference* retention times of LaMas and
+#'     *observed* retention times of matching peaks in the same sample from
+#'     which the retention times in `rt_raw` are.
+#'
+#' @author Carl Brunius
```
jorainer: add yourself as an author here too

R/do_adjustRtime-functions.R
```diff
@@ -582,3 +582,283 @@ adjustRtimeSubset <- function(rtraw, rtadj, subset,
...
+    rt_map <- rbind(c(0,0), rt_map)
+    weights <- rep(1, nrow(rt_map))
+    weights[1L] <- zero_weight
+
+    if (method == "gam") {
+        .check_gam_library()
+        model <- mgcv::gam(ref ~ s(obs, bs = bs), weights = weights,
+                           data = rt_map)
+    } else {
+        cw <- tryCatch(
```
jorainer: not sure about the catching of the warnings anymore - the code looks rather clumsy now. what if we keep the code simple and clean and mention in the documentation that, if there are too few data points, not well spaced, loess and gam might throw warnings? and that in that case users might want to reconsider/re-evaluate their lamas and matching to chrom peaks?

R/do_adjustRtime-functions.R
```diff
@@ -582,3 +582,283 @@ adjustRtimeSubset <- function(rtraw, rtadj, subset,
...
+                message("The Loess Warnings are: ", cw)
+            }
+    }
+    ## compute outliers
+    SSq <- resid(model)^2
+    meanSSq <- mean(SSq)
+    not_outlier <- (SSq / meanSSq) < resid_ratio
+
+    ## re-run only if there is outliers and keep the zero.
+    if (sum(!not_outlier)){
```
jorainer: maybe use here `if (any(!not_outlier)) {` instead of `sum` - `any` will loop through the logical vector and break the first time it finds a `TRUE` while `sum` will iterate until the end - it's just a very minor performance difference here, but it's good to keep such things in mind too.

R/do_adjustRtime-functions.R
```diff
@@ -582,3 +582,283 @@ adjustRtimeSubset <- function(rtraw, rtadj, subset,
...
+        not_outlier[1] <- TRUE
+        rt_map <- rt_map[not_outlier, , drop = FALSE]
+        weights <- weights[not_outlier]
+        if (method == "gam") {
+            model <- mgcv::gam(ref ~ s(obs, bs = "tp"), weights = weights,
+                               data = rt_map)
+        } else {
+            suppressWarnings(model <- loess(ref ~ obs, data = rt_map, span = span,
```
jorainer: I would maybe also drop the `suppressWarnings` here and *expose* the user to all warnings and hence the full amount of potential issues with the provided data.

R/do_adjustRtime-functions.R
```diff
@@ -582,3 +582,283 @@ adjustRtimeSubset <- function(rtraw, rtadj, subset,
...
+#' is.unsorted(x, na.rm = TRUE)
+#'
+#' @noRd
+.force_sorted <- function(x){
+    # Select only the non-NA values
+    nna_idx <- which(!is.na(x))
+    vec_temp <- x[nna_idx]
+
+    while (any(diff(vec_temp) < 0)) {
+        # had to change because which.max gives 1 if the vector is all FALSE..
```
jorainer: ah, that's interesting.

jorainer: an alternative would also be `idx <- match(TRUE, diff(vec_temp) < 0)` , but that is slightly slower.

R/do_adjustRtime-functions.R
```diff
@@ -582,3 +582,283 @@ adjustRtimeSubset <- function(rtraw, rtadj, subset,
...
+        idx <- which.max(diff(vec_temp) < 0)
+        # Find next biggest value
+        next_idx <- suppressWarnings(min(which(vec_temp > vec_temp[idx])))
```
jorainer: assuming that `vec_temp` is already sorted (up to `idx`) we could also simply select the first index, which avoids the `min` call and the `suppressWarnings`:

```r
next_idx <- which(vec_temp > vec_temp[idx])[1L]
```

`next_idx` will be `NA` if there is no larger value (i.e. `integer()[1L]` is `NA`). You could then below simply ask for `if (is.na(next_idx)) {...`

R/functions-Params.R
```diff
@@ -272,6 +274,52 @@ PeakGroupsParam <- function(minFraction = 0.9, extraPeaks = 1,
...
+        zeroWeight = zeroWeight,
+        ppm = ppm,
+        tolerance = tolerance,
+        toleranceRt = toleranceRt,
+        bs = bs,
+        rtMap = rtMap,
+        nChromPeaks = nChromPeaks)
+}
+
+#' Function to access rtMap from `LamaParama` object
```
jorainer: Am not completely happy with the name `rtMap` - it's OK if we name the slot like that, but to the user that might be completely unclear what we mean with that. A maybe more descriptive name might be better? ... didn't find anything though ... but then we should at least call it `rtimeMap`?

vignettes/xcms.Rmd
```diff
@@ -1285,6 +1284,141 @@ identification of e.g. features with significant different
...
+can involve comparing runs of the same samples conducted across different
+laboratories or runs with MS2 recorded after the initial MS1 run. Across
+laboratories and over time, the same samples may result in variation in
+retention time, especially because the LC system can be quite unstable. In these
+cases, an alignment step using the `adjustRtime()` function with the
+`LamaParam` parameter can allow the user to perform this type of alignment.
+We will go through this step by step below. 
+
+Let's load an already analyzed dataset `ref` and our previous dataset before
+alignment, which will be `tst`. We will first restrict their retention time
```
jorainer, line 1299: I'm wondering - do we need to have the same retention time range? I think we just did that for the unit test? if we do that here it sounds like this is a prerequisite.

vignettes/xcms.Rmd
```diff
@@ -1285,6 +1284,141 @@ identification of e.g. features with significant different
...
+range(rtime(ref))
+
+# Adjust so that they have the same.
+tst <- filterRt(tst, rt = c(2547, 4265))
+ref <- filterRt(ref, c(2547, 4265))
+```
+
+Now, we will try to align these two samples to the previous dataset. The first
+step is to extract landmark features (called `lamas`). For this, we will
+determine the features that are present in every sample fthe `ref` dataset
```
jorainer: *fthe* -> *of the*

vignettes/xcms.Rmd
```diff
@@ -1285,6 +1284,141 @@ identification of e.g. features with significant different
...
+f[f == "QC"] <- NA
+ref <- filterFeatures(ref, PercentMissingFilter(threshold = 0, f = f))
+ref_mz_rt <- featureDefinitions(ref)[, c("mzmed","rtmed")]
+ref_mz_rt
+```
+
+This is what the `lamas` input should look like for alignment. In terms of
+how this method works, the alignment algorithm matches chromatographic peaks
+from the experimental data to the lamas, fitting a model based on this
+match to adjust their retention times and minimize discrepancies between
```
jorainer: maybe instead of *discrepancies* say *differences*?

jorainer (changes requested):
Looks great Phili! I have some general suggestions on restructuring the documentation - having all documentation related to the LaMa method in one place. The code and the vignette looks however great.

R/AllGenerics.R
philouail: Do i keep some parameter in the `adjustRtime()` documentation of I can put them all only in the LamaParama documentation ?

R/do_adjustRtime-functions.R
philouail: I would agree with that too i'm not very happy with how it looks. Especially if we prevent model to be build with less than 10 rows in the rtMap. Do you agree with that btw (see above) ?

R/XcmsExperiment.R
```diff
@@ -1357,6 +1357,68 @@ setMethod(
...
+        if(length(param@rtMap) == 0)
+            param <- matchLamasChromPeaks(object, param)
+        rtMap <- param@rtMap
+        if(length(rtMap) != length(object))
+            stop("Mismatch between the number of files matched to lamas: ",
+                 length(rtMap), "and files in the object: ", length(object))
+
+        # Make model and adjust retention for each file
+        rt_adj <- bpmapply(rtMap, rt_raw, idx, FUN = function(x, y, i, param) {
+            if (nrow(x) >= 10) { # too strict ? Gam always throws error when less than that and loess does not work that well either.
```
philouail, line 1387: This is the check where I'm not completely sure (see my comment below). COuld be very dataset specific. This is what I decided based on my test datasets results.

R/do_adjustRtime-functions.R
philouail: Oh that's great I was wondering how I would manage a problem like that ! thanks

R/do_adjustRtime-functions.R
philouail: Yes I thought about `match()` but it gives NA when the vector is ordered so i can't have it as the `while` condition. and outside of the condition statement i think `which.max` is better ?

R/do_adjustRtime-functions.R
philouail: Oh wow of course !! thanks

R/functions-Params.R
philouail: You are right, what about sometimes with the word `match` or `matches` . 
Something like or `lamaMatches`, `lamaChrompeakMatches` or `matchedLamaChrompeaks`, `matchedRtime` to mimic the metaboAnnotation namings ?

R/AllGenerics.R
jorainer: I would move all over to the LamaParama documentation. only keep the mention here that `param` can also be a `[LamaParama()]`.

R/do_adjustRtime-functions.R
jorainer: then let's remove the suppression of warnings.

and yes, I'm OK for now with the hard cut of 10 - but keep the message there. would be good to hear Carl's thoughts on that.

R/do_adjustRtime-functions.R
jorainer: yes, I think your approach is better.

R/functions-Params.R
jorainer: so, we call `matchLamasChromPeaks` to create this result. the result is a list of matched retention times between lamas and chrom peaks. Somehow the `matchedRtimes` would work maybe best? look and feel would be:

```r
lapa <- LamaParama()
lapa <- matchLamasChromPeaks(xmse, lapa)
matchedRtimes(lapa)
```

would that be OKish?

R/XcmsExperiment.R
jorainer: yep - but here we might also specifically ask Carl for feedback/thoughts. I would be OK with this, but we need then to write in the documentation that we require at least 10 matched lama-chrom peaks

R/functions-Params.R
philouail: That sounds good to me !

R/XcmsExperiment.R
```diff
@@ -1357,6 +1357,66 @@ setMethod(
...
+        rt_raw <- split(rtime(object), fidx)
+        idx <- seq_along(object)
+
+        # Check if user as ran matching lama vs chrompeaks beforehand
+        if (length(param@rtMap) == 0)
+            param <- matchLamasChromPeaks(object, param)
+        rtMap <- param@rtMap
+        if (length(rtMap) != length(object))
+            stop("Mismatch between the number of files matched to lamas: ",
+                 length(rtMap), "and files in the object: ", length(object))
```
jorainer: Please add here a whitespace before `"and files...` -> `" and files...`

R/do_adjustRtime-functions.R
```diff
@@ -582,3 +582,418 @@ adjustRtimeSubset <- function(rtraw, rtadj, subset,
...
+###### LamaParama
+
+#' @title Landmark-based alignment: aligning a dataset against an external
+#' reference
+#'
+#' @aliases LamaParama-class
+#'
+#' @description
+#' Alignment is achieved using the ['adjustRtime()'] method with a `param` of
+#' class `LamaParama`. This method #' corrects retention time by aligning
```
jorainer: there's an extra `#'` in the middle of the line - please remove

R/do_adjustRtime-functions.R
```diff
@@ -582,3 +582,418 @@ adjustRtimeSubset <- function(rtraw, rtadj, subset,
...
+#' chromatographic data with an external reference dataset. Peaks in the
+#' experimental chromatographic data are aligned to predefined landmark
```
jorainer: maybe separate here the matching from the alignment?

*... dataset. Chromatographic peaks in the experimental data are first matched to predefined (external) landmark features based on their mass-to-charge ratio and retention time and subsequently the data is aligned by minimizing the differences in retention times between the matched chromatographic peaks and lamas.*

R/do_adjustRtime-functions.R
```diff
@@ -582,3 +582,418 @@ adjustRtimeSubset <- function(rtraw, rtadj, subset,
...
+#' features called "lamas" based on their mass-to-charge ratio (m/z) and
+#' retention time.
+#'
+#' This alignment algorithm matches chromatographic peaks from the experimental
```
jorainer: Then you could say here: *In detail, this...*

R/do_adjustRtime-functions.R
```diff
@@ -582,3 +582,418 @@ adjustRtimeSubset <- function(rtraw, rtadj, subset,
...
+#' data to the lamas, fitting a model based on this match to adjust their
+#' retention times and minimize discrepancies. This adjustment is performed
+#' file by file. Adjustable parameters such as `ppm`, `tolerance`, and
+#' `toleranceRt` define acceptable deviations during the matching process.
+#' It's crucial to note that only lamas and chromatographic peaks exhibiting a
+#' one-to-one mapping are considered when estimating retention time shifts. If
+#' a file has no peaks matching with lamas, no adjustment will be performed,
+#' and the file will be returned as-is. Users can evaluate this matching, for
```
jorainer: I would rather say *the retention times will be returned as-is* instead of *the file will be returned as-is*.

R/do_adjustRtime-functions.R
```diff
@@ -582,3 +582,418 @@ adjustRtimeSubset <- function(rtraw, rtadj, subset,
...
+#' reference data points and observed matching ChromPeaks. Additional
+#' parameters such as `span`, `weight`, `outlierTolerance`, `zeroWeight`,
+#' and `bs` are specific to these models. These parameters offer flexibility
+#' in fine-tuning how the matching chromatographic peaks are fitted to the
+#' lamas, thereby generating a model to align the overall retention time for
+#' a single file.
+#'
+#' Other functions related to this method:
+#'
+#'  - `LamaParama()`: return the respective the parameter object for alignment
```
jorainer: *... the respective the parameter ...* -> *... the respective parameter...*

jorainer (changes requested):
Really great Phili! only very little minor comments. After fixing that please convert to a PR and re-request review, and also review from Steffen.

R/do_adjustRtime-functions.R
philouail: Mmm i prefer the way you wrote it but it was very repetitive with the text below so i reorganised a bit, can you tell me if this update is okay for you ? it also reduce the text a bit (which i think is good)

vignettes/xcms.Rmd
```diff
@@ -1285,6 +1284,134 @@ identification of e.g. features with significant different
...
+tst <- loadXcmsData("faahko_sub2")
+```
+
+Now, we will try to align these two samples to the previous dataset. The first
+step is to extract landmark features (called `lamas`). For this, we will
+determine the features that are present in every sample of the `ref` dataset
+using the `PercentMissingFilter()` parameter in the `filterFeatures()` function.
+
+```{r}
+f <- sampleData(ref)$sample_type
```
jorainer, line 1317: maybe this definition of `f` might be a bit misleading/confusing if not explained? If you say that we want to identify features that are present in **every** sample of `ref`, that it would maybe be better to use `f <- rep(1, length(ref))`? with your code you are selecting features that are present in all samples, except QC samples.

R/do_adjustRtime-functions.R
jorainer: yes, the new documentation is perfect!

jorainer (approved):
Thanks Phili for this great contribution!

I have only one very minor comment in the vignette.

vignettes/xcms.Rmd
philouail: Actually for this alignment technique purpose, should we encourage people to maybe base it on features present in all QCs samples ? 
´f <- rep(1, length(ref))` works perfectly for this example but no real experiment will be like that with QCs and one sample type only no ?

DESCRIPTION
```diff
@@ -79,7 +79,8 @@ Suggests:
     RANN,
     multtest,
     MsBackendMgf,
-    signal
+    signal,
+    mgcv
```
sneumann, line 87: Hi, I see a new package dependency here. Where is it used ? 

I see an explicit `library(mgcv)` in `test_do_adjustRtime-functions.R`,
but I suspect it us used somewhere in `xcms/R/*.R` as well. 

I'd love to be able to handle `DESCRIPTION` and `NAMESPACE` through roxygen2 one day, and for that we'd need to start annotations 
like `#' @importFrom MsCoreUtils ndotproduct`  
Does that make sense for https://github.com/sneumann/xcms/blob/805ac3f8ba4499d41068eb2748c3ff5a4dbc752a/R/functions-Params.R#L278 ?

Or is the package really only needed in the `test_do_adjustRtime-functions.R` ?

philouail: So in `do_adjustRtime-functions.R` you will see there is helper function a function called `.check_gam_library`  which will check if `mgcv` is installed and that is only necessary if the user decide to run a `gam` model (e.g. `method = "gam"`, default being `method = "loess"`). 
The function  `mgcv:::gam()` in this package is then used in the `.rt_model()` helper function (again only if the user specifically ask for it).   
This was a suggestion from @jorainer, if i understood well that allows for the package to only be in `suggested` and therefore for xcms to not be fully dependent on it, which would not be ideal as only this implementation uses it. Please correct if i'm wrong as I am not 100% confident about that. 
I do agree that it would be great if xcms namespace could be regulated by roxygen2, for the other implementations I've done I added these `@importFrom` mentions (e.g. `filterFeatures.R`). 

Hope that was clear :) 

ps: However leaving the `library(mgcv)` in `test_do_adjustedRtime-function.R` was a mistake on my part, I forgot to remove it from when I was making the test. I will remove that in the next push !

jorainer: Also chiming in here - I absolutely want to avoid any additional (potentially unnecessary) dependency. We can have the `mgcv` either in Suggests or Enhances since we're **not** importing any functionality. We're loading the namespace specifically in the function, and only if that method is used. If the package is not installed an error is thrown that, to use the functionality, the `mgcv` package needs to be installed.

jorainer: Yes, maybe it's good (also good practice in general) to already add the `@importFrom` roxygen directives into the documentation. @philouail , can you please add them to your new functions?

vignettes/xcms.Rmd
jorainer: indeed - but then I would maybe a) describe that, and b) use just the QCs and not the other samples (at present you use just the study samples and exclude the QCs).

philouail: Yes okay sounds good i'll adjust that !

DESCRIPTION
philouail: I think it should be good now, just had to move one package around to make it more clear + added the now published `force_sorted()`, function from `MsCoreUtils`.

sneumann:
Excellent, so let's merge now and watch the build reports. Yours, Steffen

Issue #739, opened 2024-04-11
Waters data problems with Obiwarp


parasitetwin:
Hello!

So we've been working with Waters data in our lab since half a year back and want to use xcms for processing.
We have, however, run into some trouble with Obiwarp when we're trying to use the "adjustRTime()" function on the data.
I think that I have been able to find the source of the problem but I'm not sure how to alleviate it on my own.

Waters ".raw" files always (at least for the two instruments we've been working with; SynaptXS and MRTs) contain different channels for MS1 fullscan data (channel 1) and for the lock mass (channel 2).
When converting to ".mzML" format in MSConvert we remove the lock mass channel by using the subset option and only keeping the first channel (MS1 fullscan data).
Using xcms 3 the following error was received for every file:
```
Aligning 2024-02-03_B1W49_RP_NEG_sQC07_172.mzML against 2024-02-03_B1W49_RP_NEG_sQC03_128.mzML ... Found gaps in scan times of the center sample: cut scantime-vector at 29.95886743068 seconds.
Found gaps in scan time of file 2024-02-03_B1W49_RP_NEG_sQC07_172.mzML: cut scantime-vector at 29.9558508396 seconds.
OK
```
The lockmass signal is measured every 30 seconds which is what lead me to believe that the gap created by the measurement on the second channel is what is causing the problem.

In xcms 4 the problem seems to persist. I base this on the plot received when running this code:
```
group_colors <- paste0(brewer.pal(3, "Set1")[1:2], "60")
sample_colors <- group_colors[as.factor(xcmsData@sampleData$sample_group)]
par(mfrow = c(3, 1), mar = c(4.5, 4.2, 1, 0.5))
plot(bpis, col = sample_colors)
grid()
plot(bpis_adj, col = sample_colors)
grid()
## Plot also the difference of adjusted to raw retention time.
plotAdjustedRtime(xcmsData_rtAdj, col = sample_colors)
grid()
```
![image](https://github.com/sneumann/xcms/assets/23104530/69279006-ec41-4bfe-88ba-ae6d3389b6e1)
As can be seen in the picture the scantime vector does seem to be cut at ~30 seconds since all the warping is done previous to that point.

In xcms 4 running "adjustRTime()" didn't generate any warnings (I realize this can be due to my settings and setup hehe) but based on the warning from xcms 3 I dug into the code to find the source of the problem and found this section:
```
## 1)Check the scan times of both objects:
        scantime1 <- unname(rtime(cntr))
[...]

PR #740, opened 2024-04-16
fix: add parameter rtimeDifferenceThreshold for ObiwarpParam

jorainer:
- Add a parameter `rtimeDifferenceThreshold` for `ObiwarpParam` to allow increasing the tolerance to define *gaps* in the sequence of retention times. This fixes issue #739.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: xcms
-Version: 4.1.12
+Version: 4.1.13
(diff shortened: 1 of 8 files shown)


jorainer:
@sneumann , would be great if we could get this into the devel branch and onto Bioc devel before the 3.19 release.

sneumann (approved):
Excellent, thanks to the initial reporter, and Jo for the fix. Yours, Steffen

Issue #734, opened 2024-03-26
Error for XIC plot of timsTof data


NorStorz:
Hi,
i run into the following error when i try to call `plot(mse, type = "XIC")` on timfTOF data with MsBackendTimsTof:
 
`Error in rep(rtime(z), lns) : invalid 'times' argument`
 
This example should be reproducible:
```
library(xcms)
library(MsExperiment)
library(opentimsr)
library(MsBackendTimsTof)
 
so_folder <- tempdir()
so_file <- download_bruker_proprietary_code(so_folder, method = "wget")
setup_bruker_so(so_file)
 
link <- "[https://www.biosino.org/download/node/data/public/OED788908"](https://www.biosino.org/download/node/data/public/OED788908%22)
download.file(link ,destfile="dat.tar.gz")
untar("dat.tar.gz")
 
fls <- "12_serial_diluted_urine_neg/1X/04_1X_nist_urine_neg_6_p1-a1_1_7154.d"
 
mse <- readMsExperiment(fls, source = MsBackendTimsTof())
 
plot(mse, type = "XIC")
```
 
From the traceback i could find out that the error occurs inside the function  `.xmse_plot_xic ` 
I tested it with the following code: 
```
msLevel = 1L
z <- mse[1]
lst <- as(filterMsLevel(spectra(z), msLevel = msLevel), "list")
lns <- lengths(lst) / 2
lst <- do.call(rbind, lst)
df <- data.frame(rt = rep(rtime(z), lns), lst)
```
The problem seems to be that `rtime(z)` and 'lns' differ in length. This is probably caused by the MS1 filtering of the Spectra:
```
> length(rtime(z))
[1] 1839591
> length(spectra(z))
[1] 1839591
> length(lns)
[1] 1260153
> length(as(filterMsLevel(spectra(z), msLevel = 1L), "list"))
[1] 1260153
```
 
This is my session info:
```
R version 4.3.2 (2023-10-31)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 22.04.3 LTS
 
Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3 
LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.20.so;  LAPACK version 3.10.0
 
locale:
[1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
[...]

PR #741, opened 2024-04-19
fix: filtering MsLevel for xic plot

NorStorz:
-  fix for iusse #734 
- `df` is now constructed using filtered data (`flt`)
- added test for plot with ms2 data

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: xcms
-Version: 4.1.13
+Version: 4.1.14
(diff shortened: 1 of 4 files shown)


jorainer (changes requested):
Looks good, thanks @NorStorz ! Can you please also bump the version in the DESCRIPTION and add an entry in NEWS.md describing the change?

sneumann:
Thanks @NorStorz , looks good now. Yours, Steffen

PR #749, opened 2024-05-30
refactor: add storeResults/loadResults for Spectra etc

jorainer:
Some thoughts on how to maybe implement the text import/export in a more future-proof (and modular) way.


philouail (approved):
thanks for the notes and ideas !! I'll merge and start working on that

PR #753, opened 2024-06-06
refactor: updates of documentation and extension of unit tests

jorainer:
- Small fixes and formatting changes in the code.
- Expand unit tests to cover additional scenarios.
- Update/restructure documentation.

### DIFF
--- a/R/PlainTextParam.R
+++ b/R/PlainTextParam.R
@@ -1,4 +1,5 @@
 #' @include XcmsExperiment.R
+#'
 #' @title Store contents of `MsExperiment` and `XcmsExperiment` objects as
 #' plain text files
 #'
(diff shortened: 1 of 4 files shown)


R/PlainTextParam.R
```diff
@@ -9,76 +10,79 @@
...
-#' and will be exported, the full raw MS data (of the object's `Spectra` object)
-#' will currently not be exported in plain text format. For `Spectra` using the
-#' [MsBackendMzR()] backend, the names of the raw data files will however be
-#' exported (which enables to *load* the full `Spectra` respectively
-#' `MsExperiment` objects).
-#'
-#' For an `MsExperiment` object, the exported files include:
 #'
-#' - The [sampleData()] stored as a text file named *sample_data.txt*.
+#' The `storeResults()` and `loadResults()` methods with the `PlainTextParam`
```
jorainer, line 14: I restructured the documentation a bit, focusing more on the `MsExperiment` and `XcmsExperiment` and moving the ones on `Spectra` and `MsBackendMzR` a bit down - also to make it easier to eventually later move these methods to the *Spectra* package.

R/PlainTextParam.R
```diff
@@ -177,38 +182,36 @@ setMethod("storeResults",
 setMethod("loadResults",
           signature(object = "MsExperiment",
                     param = "PlainTextParam"),
-          function(object, param, spectraPath = character()){
+          function(object, param, spectraPath = character()) {
               fl <- file.path(param@path, "sample_data.txt")
-              if (file.exists(fl)){ # should i have a error if does not exist ?
```
jorainer, line 182: yes, I would check if the file exists and throw an error if not.

R/PlainTextParam.R
```diff
@@ -177,38 +182,36 @@ setMethod("storeResults",
...
-                  sd <- read.table(fl)
-                  rownames(sd) <- NULL #read.table force numbering of rownames
-              }
-              s <- loadResults(Spectra(), param)
+              if (!file.exists(fl))
+                  stop("No 'sample_data.txt' file found in the provided path.")
+              sd <- read.table(fl, sep = "\t")
+              rownames(sd) <- NULL #read.table force numbering of rownames
+              s <- loadResults(Spectra(), param, spectraPath = spectraPath)
```
jorainer, line 191: and we need to pass the `spectraPath` to all `loadResults` methods

R/PlainTextParam.R
```diff
@@ -177,38 +182,36 @@ setMethod("storeResults",
...
-              }
-)
+          })
 
 #' @rdname PlainTextParam
 setMethod("loadResults",
           signature(object = "XcmsExperiment",
                     param = "PlainTextParam"),
-          function(object, param, spectraPath){
-              res <- callNextMethod() #check if need to add spectraPath = spectraPath
```
jorainer, line 198: it is possible to use either `callNextMethod(object = object, spectraPath = spectraPath)`, or simply `callNextMethod()` in which case all parameters passed to the original call will be passed along to the next method.

R/PlainTextParam.R
```diff
@@ -177,38 +182,36 @@ setMethod("storeResults",
...
+          })
 
 #' @rdname PlainTextParam
 setMethod("storeResults", signature(object = "Spectra",
                                     param = "PlainTextParam"),
           function(object, param) {
               dir.create(path = param@path,
                          recursive = TRUE,
-                         showWarnings = TRUE)
+                         showWarnings = FALSE)
```
jorainer, line 214: I would disable the warnings, because a `storeResults,Spectra` will create the directory, save data there and then call `storeSpectra,MsBackendMzR` which will throw a warning because the directory was already created. So, we get these warnings even in a typical workflow - which might confuse users.

R/PlainTextParam.R
```diff
@@ -234,9 +237,9 @@ setMethod("loadResults", signature(object = "Spectra",
                                                   "PlainTextParam")))
                   stop("Can not store a 'Spectra' object with backend '",
                        variables["backend"], "'")
-              b <- loadResults(object= do.call(what = variables[["backend"]],
-                                               args = list()),
-                          param = param, spectraPath = spectraPath) ##better way to do this ?
```
jorainer, line 239: I don't know a better way.

R/PlainTextParam.R
```diff
@@ -249,37 +252,40 @@ setMethod("loadResults", signature(object = "Spectra",
...
               rownames(data) <- NULL
               data <- DataFrame(data)
               b@spectraData <- data
-              if (length(spectraPath) > 0){
-                  old <- MsCoreUtils::common_path(dataStorage(b))
-                  if (nchar(old) > 0)
-                      old <- paste0(old, "/")
+              if (length(spectraPath) > 0) {
+                  old <- common_path(dataStorage(b))
+                  ## if (nchar(old) > 0)
```
jorainer, line 287: not sure if we need this - I would skip since `/` and `//` work both, so better to be sure that there is a `/` before the file name.

R/PlainTextParam.R
```diff
@@ -367,19 +372,20 @@ setMethod("loadResults", signature(object = "MsBackendMzR",
 #' Chromatographic peaks
 #' @noRd
 .export_chrom_peaks <- function(x, path = character()) {
-    write.table(chromPeaks(x), file = file.path(path, "chrom_peaks.txt"))
-    write.table(as.data.frame(chromPeakData(x)),
+    write.table(chromPeaks(x), file = file.path(path, "chrom_peaks.txt"),
```
jorainer, line 375: I added `sep = "\t"` to all `write.table`/`read.table`. The default `sep = " "` might be problematic if there are white spaces in the file names or similar.

tests/testthat/test_PlainTextParam.R
```diff
@@ -12,42 +12,80 @@ test_that("storeResults,loadResults,PlainTextParam,MsBackendMzR works", {
     ## Loading data again
     b2 <- loadResults(object = MsBackendMzR(), param)
     expect_true(inherits(b2, "MsBackendMzR"))
-    b <- dropNaSpectraVariables(b) #the function does this to be robust, is it a problem ? i should mention it in the doc
```
jorainer, line 15: yes, that's a good point. We should mention this. but this will be specific for the `storeResults,MsBackendMzR` - for other backends we might not need that.

tests/testthat/test_PlainTextParam.R
```diff
@@ -12,42 +12,80 @@ test_that("storeResults,loadResults,PlainTextParam,MsBackendMzR works", {
...
+    b <- dropNaSpectraVariables(b)
     expect_equal(b@spectraData, b2@spectraData)
-    expect_equal(peaksVariables(b), peaksVariables(b2))  # true even without forcing the slot
+    expect_equal(peaksVariables(b), peaksVariables(b2))
+    expect_equal(mz(b[1:20]), mz(b2[1:20]))
+
+    ## Check the spectraPath parameter.
```
jorainer, line 20: this is to simulate moving a stored results to a different computer and the usage of the `spectraPath` parameter.

tests/testthat/test_PlainTextParam.R
```diff
@@ -12,42 +12,80 @@ test_that("storeResults,loadResults,PlainTextParam,MsBackendMzR works", {
...
 })
 
 test_that("storeResults,loadResults,PlainTextParam,Spectra works", {
     pth <- file.path(tempdir(), "test1")
     param <- PlainTextParam(path = pth)
-    #add processingQueueVariables to test export
+    ## add processingQueueVariables to test export
+    s@processingQueueVariables <- c(s@processingQueueVariables, "rtime")
     s <- filterMzRange(s, c(200,300))
+    s <- filterRt(s, c(3000, 3500)) # to ensure subsetted object would work
```
jorainer, line 42: this is to ensure that not only the processing queue is correctly exported/imported, but also eventual data subsets.

jorainer:
really big thanks for the great code and contribution @philouail !

R/PlainTextParam.R
philouail: Aaah thanks for the explanation

R/PlainTextParam.R
philouail: mmm I need to test it, because I thought the result  of common_path never resulted with a `/` at the end of the path, so i thought that was necessary

R/PlainTextParam.R
philouail: Ah of course ! it was working  fine so I thought it was okay , but i should keep in mind that it's better to be safe than sorry :)

tests/testthat/test_PlainTextParam.R
philouail: Aaah sorry I misunderstood that

tests/testthat/test_PlainTextParam.R
```diff
@@ -82,16 +137,39 @@ test_that("storeResults,loadResults,PlainTextParam,XcmsExperiment works", {
...
+    expect_equal(tmp@chromPeaks, load_xmse@chromPeaks)
+    expect_equal(tmp@chromPeakData, load_xmse@chromPeakData)
+    expect_equal(tmp@sampleData, load_xmse@sampleData)
+    expect_equal(length(tmp@processHistory), length(load_xmse@processHistory))
+    expect_equal(tmp@processHistory[[1L]], load_xmse@processHistory[[1L]])
+    expect_equal(tmp@processHistory[[2L]], load_xmse@processHistory[[2L]])
+    expect_equal(tmp@processHistory[[3L]], load_xmse@processHistory[[3L]])
+    expect_equal(tmp@processHistory[[4L]], load_xmse@processHistory[[4L]])
+    expect_equal(tmp@processHistory[[5L]], load_xmse@processHistory[[5L]])
+    ## The 6th param object contains functions for which the comparison fails
```
philouail, line 158: damn   it, i tried to do that but obviously only tested it on the 6th param and gave up when I saw the error :')

tests/testthat/test_PlainTextParam.R
```diff
@@ -56,21 +94,38 @@ test_that("storeResults,loadResults,PlainTextParam,MsExperiment works", {
     ## Loading data again
     load_mse <- loadResults(object = MsExperiment(), param)
     expect_true(inherits(load_mse, "MsExperiment"))
-    expect_equal(sampleData(mse), sampleData(load_mse))
-    a <- spectra(mse)
+    expect_equal(sampleData(tmp), sampleData(load_mse))
```
philouail, line 97: Can I ask why you create a `tmp`  object here ? is it because otherwise we permanently change the `mse` (or `xmse_full`) object for the later tests ?

philouail:
Thanks so much @jorainer  for polishing everything ! it looks so much  better, I do have some questions in comments !

Also the checks don't pass because of the Spectra package update aha

R/PlainTextParam.R
philouail: After checking, isn't that how the function behave ? and we definitely need to add this `/` ?

jorainer: how I understood it (can be wrong), but yes, the result from `common_path()` never ends in a `"/"`. so, assume we have `"/some/path/some/file.txt"`, then the common path would be `"/some/path/some" and assuming the user defines `spectraPath = "/new/path"`. We would like to replace to get `"/new/path/file.txt"`. If we always add a `"/"` at the `old`, `old` would be in this case `"/some/path/some/"` and the `sub()` call will replace `"/some/path/some/"` with `"/new/path"` - and we get `"/new/pathfile.txt"`. that's why I would not append the trailing `"/"` to the pattern we want to replace. 

Also, `"/new/path/file.txt"` and `"/new/path//file.txt"` both work - so, if the user would provide `spectraPath = "/new/path/"` (i.e. with a trailing `"/"`) it would/should still work.

tests/testthat/test_PlainTextParam.R
jorainer: I think `testthat` changes the variables within each unit test locally, not globally - it's more that when I evaluate the unit tests line by line I change the variables in my R session and then I get unexpected errors later. I generally avoid overwriting the "globally" defined variables where possible. Also, I specifically use a variable name `tmp` to remember that I will overwrite this variable for sure later...

jorainer:
The updates in the *Spectra* package should become available later today or tomorrow I guess - I pushed yesterday to Bioconductor.

tests/testthat/test_PlainTextParam.R
philouail: oki fair thanks !

R/PlainTextParam.R
philouail: ooooh  of course  ! No that make complete sense, but ye I think that the less we have `if` statement the better so i will remove it.

PR #758, opened 2024-07-18
small fixes

philouail:
This is all i could find from my branch that was not related to the import/export stuff :)


jorainer (approved):
Thanks for the quick fixes!

Issue #755, opened 2024-06-24
Chromatograms of MS1 data with two different scan events per file


breidan:
Hi,
this is a follow on of issue #723. @jorainer suggested to assign `msLevel` 1 to the first scan event, and 2 to the second. That way one `MsExperiment` object can be processed with `findChromPeaks(object,msLevel=1L,...)` and `findChromPeaks(object,msLevel=2L,...)` to obtain all `chromPeaks` in one object. That works fine. What is not working is extracting chromatograms with `chromatogram(object,msLevel=2L)`.

Data from an Orbitrap Elite were converted with `msconvert` to mzML and read in and then split by the two scan events (they occur in the order below):

>  msexp <- readMsExperiment(spectraFiles = files[timeord], backend = MsBackendMzR(),sampleData = pd[timeord,])
>  mse1 <- filterSpectra(msexp, filter = filterFilterString, value = "FTMS + c ESI sid=10.00 SIM ms [300.00-800.00]")
>  mse2 <- filterSpectra(msexp, filter = filterFilterString, value = "FTMS + c ESI sid=10.00 SIM ms [600.00-1200.00]")

From here I obtained the expected chromatograms with

>  chromatogram(mse1,mz=c(394.75,394.77))|>plot()
>  chromatogram(mse2,mz=c(663.70,663.72))|>plot()

If I change the `msLevel` in `msexp` for the second scan event to 2 with

> mslevel<-spectra(msexp)$msLevel
scanType<-spectra(msexp)$filterString
mslevel[grep("600.00",scanType)]<-2L
spectra(msexp)$msLevel<-mslevel

then `chromatogram(msexp,mz=(394.75,394.77))|>plot()` produces a chromatogram while `chromatogram(msexp,mz=c(663.70,663.72),msLevel=2L)|>plot()` comes up empty. In `chromPeaks` all possible peaks in the range from m/z 300 to 1200 are listed. Any ideas how I can draw the chromatograms for the second scan event? Or how to better deal with multiple scan events per file?

The versions of the relevant packages:
> [1]  Spectra_1.14.0                S4Vectors_0.42.0              BiocGenerics_0.50.0          
 [5] MsExperiment_1.6.0            ProtGenerics_1.36.0           xcms_4.2.1

jorainer:
This is strange. Can you maybe share one example file (with both MS1 and MS2 in it) for me to test? in case my email is johannes.rainer at eurac.edu

breidan:
@jorainer,
[...]

PR #761, opened 2024-08-06
Some updates and bug fixes

jorainer:
This PR adds:

- fix for issue #755 (`chromatogram()` call with `msLevel = 2L`); improve documentation (and vignette) for this.
- change the `estimatePrecursorIntensity()` function to a **method** and import the generic from *ProtGenerics*. This ensures we can have different implementations for `XcmsExperiment`, `XCMSnExp` and also avoid clashes with the implementation in *Spectra* for `Spectra` objects.
- add a function that allows coercion of a `XcmsExperiment` to `XCMSnExp` for backward compatibility.
- fix GHA-unit tests


.github/workflows/check-bioc.yml
```diff
@@ -124,32 +124,7 @@ jobs:
...
-          ## Enable installing XML from source if needed
-          brew install libxml2
-          echo "XML_CONFIG=/usr/local/opt/libxml2/bin/xml2-config" >> $GITHUB_ENV
-
-          ## Required to install magick as noted at
-          ## https://github.com/r-lib/usethis/commit/f1f1e0d10c1ebc75fd4c18fa7e2de4551fd9978f#diff-9bfee71065492f63457918efcd912cf2
-          brew install imagemagick@6
-
-          ## For textshaping, required by ragg, and required by pkgdown
-          brew install harfbuzz fribidi
```
sneumann, line 136: `brew install harfbuzz fribidi` seriously ? Somebody was creative there :-) Certainly very google'able...

DESCRIPTION
```diff
@@ -1,5 +1,5 @@
 Package: xcms
-Version: 4.3.2
+Version: 4.3.4
```
sneumann: Are we skipping 4.3.3 ? I couldn't find a 4.3.3 in a PR.

DESCRIPTION
```diff
@@ -54,7 +54,7 @@ Imports:
     methods,
     Biobase,
     BiocGenerics,
-    ProtGenerics (>= 1.35.4),
+    ProtGenerics (>= 1.37.1),
```
sneumann, line 57: That's needed because only the newer has a generic for `estimatePrecursorIntensity` ?

DESCRIPTION
```diff
@@ -63,7 +63,7 @@ Imports:
     MsCoreUtils (>= 1.15.5),
     MsFeatures,
     MsExperiment (>= 1.5.4),
-    Spectra (>= 1.13.7),
+    Spectra (>= 1.15.7),
```
sneumann, line 66: I Guess needed because only the newer has `estimatePrecursorIntensity`

sneumann (approved):
Looks good to me, and I learned about `@noRd` today ;-)

DESCRIPTION
jorainer: Exactly. The `estimatePrecursorIntensity` generic was added to *ProtGenerics* version 1.37.1

DESCRIPTION
jorainer: Yes, because for `MsExperiment`/`XcmsExperiment` we're using the `estimatePrecursorIntensity,Spectra` method, while for the older `XCMSnExp` object we use a different one.

DESCRIPTION
jorainer: True, my mistake. Somehow I forgot to make the 4.3.3 PR. I will revert to 4.3.3 for this PR.

jorainer: Fixed now.

PR #767, opened 2024-09-18
Add internal function to calculate beta metrics

pablovgd:


R/XcmsExperiment-functions.R
```diff
@@ -542,6 +542,53 @@
...
+#'
+#' @param rt retention time for each peak matrix.
+#'
+#' @param peakArea `matrix` defining the chrom peak area.
+#'
+#' @author Pablo Vangeenderhuysen
+#'
+#' @noRd
+.chrom_peak_beta_metrics <- function(x, rt, peakArea,
+                                           sampleIndex = integer(),
```
jorainer: can you please remove the `sampleIndex` and `cn` parameters? we will not need these. Just make sure to keep the `...`

R/XcmsExperiment-functions.R
```diff
@@ -542,6 +542,53 @@
...
+  res <- matrix(NA_real_, ncol = 2L, nrow = nrow(peakArea))
+  rownames(res) <- rownames(peakArea)
+  colnames(res) <- c("beta_cor","beta_snr")
+  for (i in seq_len(nrow(res))) {
+    rtr <- peakArea[i, c("rtmin", "rtmax")]
+    keep <- which(MsCoreUtils::between(rt, rtr))
+    if (length(keep)) {
+      xsub <- lapply(x[keep], .pmat_filter_mz,
+                     mzr = peakArea[i, c("mzmin", "mzmax")])
+      ## length of xsub is the number of spectra, the number of peaks can
```
jorainer: I would suggest to remove this comment (lines 570-572). Also I would remove the `mat` calculation (line 573; we will not need that) and the `if (nrow(mat))` (line 574 and 582). The `get_beta_values()` function checks internally that enough data points are provided (minimum is 5) and returns `NA` if not. So we don't need to run this part only if `nrow(mat)` is larger than one - and it saves us additional (unneeded calculations). Also, please remove the `rts <- rep(rt[keep], nr)` line - we don't use the `rts` variable, so also here, no need to calculate that. So, to summarize, simply keep lines 578:580.

R/XcmsExperiment-functions.R
```diff
@@ -542,6 +542,53 @@
...
+      ## however be 0 if no peak was found. Maybe we should/need to
+      ## consider adding 0 or NA intensity for those.
+      mat <- do.call(rbind, xsub)
+      if (nrow(mat)) {
+        nr <- vapply(xsub, nrow, NA_integer_)
+        ## can have 0, 1 or x values per rt; repeat rt accordingly
+        rts <- rep(rt[keep], nr)
+        res[i, c("beta_cor", "beta_snr")] <- .get_beta_values(
+          vapply(xsub[nr > 0], sum, NA_real_),
```
jorainer: This line `vapply(xsub[nr > 0], sum, NA_real_)` actually sums each of the numerical matrices in `xsub` (that have at least one row), i.e. it sums the m/z **and** intensity values. Please make sure you only sum the intensity values: `vapply(xsub[nr > 0], function(z) sum(z[, "intensity"]), NA_real_)`.

Could well be that I have the same code in my original function - if so, please correct also there.

jorainer (changes requested):
Thanks @pablovgd ! There are however some extra (unneeded) calculations/variables and also a bug. Please change according to my comments.

R/XcmsExperiment-functions.R
pablovgd: Also corrected in the original code of `.chrom_peak_intensity_centWave`

pablovgd:
I made the requested changes and pushed to my issue705 branch.

jorainer (approved):
Perfect! thanks @pablovgd . I'll merge. Next would be to implement the user-facing code.

PR #768, opened 2024-09-20
added chromPeakSummary method

pablovgd:
As discussed with @jorainer I added the `chromPeakSummary()` method which uses a new parameter class: `BetaDistributionParam`. It performs calculation of the new "beta quality metrics": `beta_cor` and `beta_snr` on an `XcmsExperiment `object for which `hasChromPeaks(object) = TRUE`.

I also added unit tests. 

Johannes, feel free to let me know what I forgot and/or still need to add.
Thanks!

### DIFF
--- a/R/XcmsExperiment.R
+++ b/R/XcmsExperiment.R
@@ -2013,3 +2013,44 @@ setMethod(
         object[i = sort(unique(file)), keepAdjustedRtime = keepAdjustedRtime,
                keepFeatures = keepFeatures, ...]
     })
+
+#' @rdname chromPeakSummary
+setMethod(
+  "chromPeakSummary",
+  signature(object = "XcmsExperiment", param = "BetaDistributionParam"),
+  function(object, param, msLevel = 1L, chunkSize = 2L, BPPARAM = bpparam()) {
+    if (length(msLevel) != 1)
+      stop("Can only perform peak metrics for one MS level at a time.")
+    if (!hasChromPeaks(object, msLevel = msLevel))
+      stop("No ChromPeaks definitions for MS level ", msLevel, " present.")
+    ## Define region to calculate metrics from for each file
+    cp <- chromPeaks(object)
+    f <- factor(cp[,"sample"],seq_along(object))
+    pal <- split.data.frame(cp[,c("mzmin","mzmax","rtmin","rtmax")],f)
+    names(pal) <- seq_along(pal)
+    ## Get integration function and other info.
+    ph <- .xmse_process_history(object, .PROCSTEP.PEAK.DETECTION,
+                                msLevel = msLevel)
+    ## Manual chunk processing because we have to split `object` and `pal`
+    idx <- seq_along(object)
+    chunks <- split(idx, ceiling(idx / chunkSize))
+    pb <- progress::progress_bar$new(format = paste0("[:bar] :current/:",
+                                           "total (:percent) in ",
+                                           ":elapsed"),
+                           total = length(chunks) + 1L, clear = FALSE)
+    pb$tick(0)
+    mzf <- "wMean"
+    res <- lapply(chunks, function(z, ...) {
+      pb$tick()
+      .xmse_integrate_chrom_peaks(
+        .subset_xcms_experiment(object, i = z, keepAdjustedRtime = TRUE,
+                                ignoreHistory = TRUE),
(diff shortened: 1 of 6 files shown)


R/XcmsExperiment.R
```diff
@@ -2013,3 +2013,44 @@ setMethod(
...
+    if (length(msLevel) != 1)
+      stop("Can only perform peak metrics for one MS level at a time.")
+    if (!hasChromPeaks(object, msLevel = msLevel))
+      stop("No ChromPeaks definitions for MS level ", msLevel, " present.")
+    ## Define region to calculate metrics from for each file
+    cp <- chromPeaks(object)
+    f <- factor(cp[,"sample"],seq_along(object))
+    pal <- split.data.frame(cp[,c("mzmin","mzmax","rtmin","rtmax")],f)
+    names(pal) <- seq_along(pal)
+    ## Get integration function and other info.
```
jorainer: you can drop these lines (2031:2034) as we don't need this.

R/XcmsExperiment.R
```diff
@@ -2013,3 +2013,44 @@ setMethod(
...
+                                msLevel = msLevel)
+    ## Manual chunk processing because we have to split `object` and `pal`
+    idx <- seq_along(object)
+    chunks <- split(idx, ceiling(idx / chunkSize))
+    pb <- progress::progress_bar$new(format = paste0("[:bar] :current/:",
+                                           "total (:percent) in ",
+                                           ":elapsed"),
+                           total = length(chunks) + 1L, clear = FALSE)
+    pb$tick(0)
+    mzf <- "wMean"
```
jorainer: also no need for this line as we don't need this variable.

R/XcmsExperiment.R
```diff
@@ -2013,3 +2013,44 @@ setMethod(
...
+    res <- lapply(chunks, function(z, ...) {
+      pb$tick()
+      .xmse_integrate_chrom_peaks(
+        .subset_xcms_experiment(object, i = z, keepAdjustedRtime = TRUE,
+                                ignoreHistory = TRUE),
+        pal = pal[z], intFun = .chrom_peak_beta_metrics, mzCenterFun = mzf,
```
jorainer: i would change here to `pal = pal[z], intFun = .chrom_peak_beta_metrics, BPPARAM = BPPARAM)`, we don't need the other parameters.

R/XcmsExperiment.R
```diff
@@ -2013,3 +2013,44 @@ setMethod(
...
+        param = BetaDistributionParam(), BPPARAM = BPPARAM)
+    })
+    res <- do.call(rbind, res)
```
jorainer, line 2047: I would suggest to directly return the result matrix (i.e. have here as last line `do.call(rbind, res)` and remove all lines below). So, essentially, calling `mat <- chromPeakSummary(object, BetaDistributionParam())` would result in `mat` being the two-column result matrix.

tests/testthat/test_XcmsExperiment.R
```diff
@@ -1452,3 +1452,11 @@ test_that("fillChromPeaks,XcmsExperiment works with verboseBetaColumns", {
     pks_fil <- chromPeaks(res)[chromPeakData(res)$is_filled, ]
     expect_true(sum(is.na(pks_fil[, "beta_cor"])) < 4)
 })
+
+test_that("chromPeakSummary,XcmsExperiment works", {
+  p <- CentWaveParam(noise = 10000, snthresh = 40, prefilter = c(3, 10000),
+                     verboseBetaColumns = FALSE)
+  xmse <- findChromPeaks(mse, param = p)
+  res <- chromPeakSummary(res,BetaDistributionParam())
+  expect_true(all(c("beta_cor", "beta_snr") %in% colnames(chromPeaks(res))))
```
jorainer: As detailed above, I would not replace the existing `chromPeaks` matrix with the metric, but have `chromPeakSummary()` returning the `numeric` `matrix` with the values of the metric.

jorainer (changes requested):
Thanks Pablo! nice contribution, but I have some change requests. Most importantly, I would suggest that the function returns the metrics and does not replace the `chromPeaks` matrix in the object.

jorainer (approved):
Looks good Pablo! Thanks!

PR #770, opened 2024-09-23
Added peak quality section to xcms vignette.

pablovgd:
Hi Johannes,

As discussed, I added a section on peak quality and the new `chromPeakSummary ` function in the vignette. 
I tried to adhere to the vignette style, please correct where needed :)
I also kept it rather short, to keep the vignette comprehensible. If you think it lacks certain essentials, let me know.

Cheers,
Pablo


jorainer (approved):
That's great! Thanks Pablo!

PR #772, opened 2024-09-30
Add chromPeakSummary function and a fix

jorainer:
This PR adds the `chromPeakSummary()` method and a first implementation to calculate the peak shape quality from @wkumler on chrom peak results (thanks to @pablovgd for contributing).

In addition, it fixes a bug in the calculation of the beta scores during gap filling.


philouail (approved):
This is great ! Also thanks for the small vignette ! The code looks good to me, thank @pablovgd, excited to incorporate it in the overall workflow and see how to integrate it with the other preprocessing steps to improve them.

PS: I think the GHA needs to be updated, i they are failing because of warnings.

R/AllGenerics.R
```diff
@@ -535,6 +535,64 @@ setGeneric("chromPeakData<-", function(object, value)
 setGeneric("chromPeakSpectra", function(object, ...)
     standardGeneric("chromPeakSpectra"))
 
+#' @title Chromatographic peak summaries
+#'
+#' @name chromPeakSummary
+#'
+#' @description
+#'
+#' The `chromPeakSummary()` method calculates summary statistic or other
```
wkumler: ce: "statistics" instead of "statistic"

R/AllGenerics.R
```diff
@@ -535,6 +535,64 @@ setGeneric("chromPeakData<-", function(object, value)
...
+#' metrics for each of the identified chromatographic peaks in an *xcms*
+#' result object, such as the [XcmsExperiment()]. Different metrics can be
+#' calculated, depending (and configured) using dedicated *parameter* classes.
```
wkumler: ce: depending upon  (and configured by) dedicated *parameter* classes.

R/AllGenerics.R
```diff
@@ -535,6 +535,64 @@ setGeneric("chromPeakData<-", function(object, value)
...
+#' As a result, the method returns a `matrix` or `data.frame` with one row
+#' per chromatographic peak. Each column contains calculated values, depending
+#' on the used method/parameter class.
+#'
+#' Currently implemented methods/parameter classes are:
+#'
+#' - `BetaDistributionParam`: calculates the *beta_cor* and *beta_snr* quality
+#'   metrics as described in (Kumler 2023) representing the result from a
```
wkumler: Maybe drop the parentheses around (Kumler 2023) or add a direct link?

R/AllGenerics.R
```diff
@@ -535,6 +535,64 @@ setGeneric("chromPeakData<-", function(object, value)
...
+#'   (correlation) test of similarity to a bell curve and the signal-to-noise
```
wkumler: test of similarity (using Pearson's correlation coefficient)

R/AllGenerics.R
```diff
@@ -535,6 +535,64 @@ setGeneric("chromPeakData<-", function(object, value)
...
+#'   ratio calculated on the residulas of this test.
```
wkumler: ce: residuals

R/XcmsExperiment-functions.R
wkumler, line 1: Looks great, thanks for implementing this! I haven't checked it for bugs but the logic looks sound.

R/methods-XChromatogram.R
```diff
@@ -28,10 +28,13 @@ setMethod("show", "XChromatogram", function(object) {
...
 #'   description for details). For `XChromatogram` objects the function returns
-#'   a `matrix` with columns `"rt"` (retention time of the peak apex),
-#'   `"rtmin"` (the lower peak boundary), `"rtmax"` (the upper peak boundary),
-#'   `"into"` (the ingegrated peak signal/area of the peak), `"maxo"` (the
-#'   maximum instensity of the peak and `"sn"` (the signal to noise ratio).
+#'   a `matrix` with columns `"mz"` (mean m/z value), `"mzmin"` (minimal m/z
+#'   value) and `"mzmax"` (maximal m/z value), `"rt"` (retention time of the
+#'   peak apex), `"rtmin"` (the lower peak boundary in retention time
+#'   dimension), `"rtmax"` (the upper peak boundary in retention time
+#'   dimension), `"into"` (the ingegrated peak signal/area of the peak),
```
wkumler: ce: integrated

vignettes/xcms.Rmd
```diff
@@ -431,6 +431,56 @@ numeric).
...
+`chromPeakChromatograms`. An example of a peak with a high *beta_cor* and for
+a peak  with a low *beta_cor* score is given below.
+
+```{r chromPeakChromatograms, message=FALSE}
+beta_metrics[c(4, 6), ]
+eics <- chromPeakChromatograms(
+    faahko, peaks = rownames(chromPeaks(faahko))[c(4, 6)])
+```
+
+```{r peak-quality-metrics, fig.widht = 10, fig.height = 5, fig.cap = "Plots of high and low quality peaks. Left: peak CP0004 with a beta_cor = 0.98, right: peak CP0006 with a beta_cor = 0.13."}
```
wkumler: fig.width instead of fig. widht? Not sure if this matters if the vignette still compiles nicely.

vignettes/xcms.Rmd
```diff
@@ -431,6 +431,56 @@ numeric).
 chromPeakData(faahko)
 ```
 
+### Chromatographic peak quality
+
+Based on the publication by Kumler et al. published in 2023 [@Kumler2023], new
+quality metrics (*beta_cor* and *beta_snr*) were added to *xcms*. *beta_cor*
+indicates how bell-shaped the chromatographic peak is and *beta_snr* is
+estimating the signal-to-noise ratio using the residuals from this fit. These
```
wkumler, line 442: Could be worth mentioning here why this metric might differ from the existing "sn" column. In my experience this mostly occurs with chromatographic peaks that don't have very long tails and there isn't enough information outside of the peak to robustly estimate noise, resulting in a hard-coded value of noise = 1. So that people could still use the sn metric if they know to be skeptical of the times when sn = floor(maxo).

jorainer:
Thanks @wkumler for your review! I have now addressed all your suggestions.

R/do_findChromPeaks-functions.R
```diff
@@ -3720,14 +3720,20 @@ peaksWithCentWave <- function(int, rt,
...
 #' @param rtime A numeric vector corresponding to the retention times of each
-#' intensity. If not provided, intensities will be assumed to be equally spaced.
+#'     intensity. Retention times are expected to be in increasing order
+#'     without duplicates. If not provided, intensities will be assumed to be
+#'     equally spaced.
+#'
 #' @param skews A numeric vector of the skews to try, corresponding to the
-#' shape1 of dbeta with a shape2 of 5. Values less than 5 will be increasingly
-#' right-skewed, while values greater than 5 will be left-skewed.
+#'     shape1 of dbeta with a shape2 of 5. Values less than 5 will be
```
sneumann, line 3730: Hi @wkumler , I needed to read that a few times. So a value of 5 means symmetric ? Why is it not zero centered, so that `abs()` tells you the skewedness (regardless in what direction) and you can `<0` or `>0` if you only want the direction ?

R/functions-XCMSnExp.R
```diff
@@ -279,6 +279,7 @@ dropGenericProcessHistory <- function(x, fun) {
                        valsPerSpect = valsPerSpect, rtrange = rtr,
                        mzrange = mzr)
         if (length(mtx)) {
+            ## mtx: time, mz, intensity
             if (any(!is.na(mtx[, 3]))) {
```
sneumann, line 283: I usually recommend avoiding hardcoded column numbers, imagine someone came up with `(time, ccs, mz, intensity)`. Might need that mtx is created with named columns somewhere above. Is that the only occurance of a hardcoded column number in the PR ?

R/functions-XCMSnExp.R
```diff
@@ -290,21 +291,20 @@ dropGenericProcessHistory <- function(x, fun) {
                 ## as e.g. centWave. Using max(1, ... to avoid getting Inf in
                 ## case the signal is based on a single data point.
                 ## (3) rtr[2] - rtr[1]
-                res[i, "into"] <- sum(mtx[, 3], na.rm = TRUE) *
+                res[i, "into"] <- sum(mtx[, 3L], na.rm = TRUE) *
                     ((rtr[2] - rtr[1]) /
                      max(1, (sum(rtim >= rtr[1] & rtim <= rtr[2]) - 1)))
-                maxi <- which.max(mtx[, 3])
+                maxi <- which.max(mtx[, 3L])
```
sneumann, line 297: No, above was not the only hardcoded `3` :-) and more below ...

sneumann (changes requested):
Hi, thanks for the PR ! 
What I couldn't attach to a specific line, would it make sense to add the paper to https://github.com/sneumann/xcms/blob/devel/inst/CITATION ?
Yours, Steffen

R/do_findChromPeaks-functions.R
wkumler: That's a good point, thank you for the review. I'm using the language inherent to the `dbeta()` function where the values of 2-5 correspond to the "positive parameters" alpha and beta (or, in R, shape1 and shape2). I'm absolutely open to rescaling these and I agree that a positive/negative skew would be more intuitive. If we're open to rescaling and we have a parameter to pass now in `chromPeakSummary` then we also may want to allow a wider variety of numbers - the values of 5 matched my intuition for peak shape corners but other folks may want more/less tail and more/less skew.

R/functions-XCMSnExp.R
jorainer: I agree - generally. Here, we load the `mtx` `matrix` with a function that is supposed to return a 3 column matrix with the columns exactly in the expected order. Thus, IMHO for this case (as long as there is no user input involved) it's save to use access-by-index. Also, because here we're calling this in a loop thousands of times, the way how we subset *could* affect the performance.

I did some timings on that:

```r
mtx <- cbind(time = 1:5, mz = 1:5, intensity = c(2342.2, 123.1, 231.1, 23.1, 123.23))
int_col <- 3L

library(microbenchmark)
> microbenchmark(mtx[, 3L], mtx[, "intensity"], mtx[, int_col])
Unit: nanoseconds
               expr min    lq   mean median    uq  max neval cld
          mtx[, 3L] 354 363.5 437.30  380.5 477.5  948   100   a
 mtx[, "intensity"] 389 404.0 543.07  412.0 474.0 5036   100   a
     mtx[, int_col] 381 399.5 469.45  407.0 484.0 1135   100   a
```

jorainer:
Regarding the CITATIONS - I added the reference to the respective man page. I think adding that as a CITATION for xcms itself might be bit too much, as the paper deals more with peak shape quality, independently of xcms?

jorainer:
For now I only bumped the version and fixed the NEWS.md, but I did not address the other points above - can we maybe discuss them again @sneumann ?

jorainer:
I'll make the PR a draft and fix the conflicts. After that I'll re-open.

jorainer:
I've now fixed the conflicts and in addition also added the `c()` method to combine multiple `XcmsExperiment` objects into one as well as method to coerce from `XCMSnExp` to `XcmsExperiment` objects. Would be ready to review @sneumann :)

jorainer:
@sneumann , just a little reminder that this PR is still open and we need to address it :)

R/XcmsExperiment-functions.R
```diff
@@ -1169,6 +1209,70 @@ XcmsExperiment <- function() {
     as(MsExperiment(), "XcmsExperiment")
 }
 
+#' Convert a XCMSnExp to a XcmsExperiment.
+#'
+#' @noRd
+.xcms_n_exp_to_xcms_experiment <- function(from) {
```
sneumann, line 1215: Unusual (to me) way of converting CamelCase class name to (part of) function name. As it is internal, I am just curious.

R/XcmsExperiment-functions.R
```diff
@@ -1169,6 +1209,70 @@ XcmsExperiment <- function() {
...
+    res
+}
+
+#' Combine `XcmsExperiment` objects. Only combining of chrom peaks is supported.
+#' Any alignment or correspondence results are removed.
+#'
+#' @param x `list` of `XcmsExperiment` objects.
+#'
+#' @noRd
+.xmse_combine <- function(x) {
```
sneumann, line 1249: Again internal, but would be `xcmse` or something else be more intiutive ?

R/XcmsExperiment.R
```diff
@@ -47,6 +47,18 @@
...
 #'   `FALSE`) and `keepFeatures` (by default `FALSE`).
 #'
+#' - `c`: multiple `XcmsExperiment` objects can be combined into one using the
+#'   `c()` function. This requires however that all the `XcmsExperiments`'
+#'   `Spectra` objects use the same type of `MsBackend` and that their
+#'   processing queues are empty. Also, only combining of peak detection
+#'   results is supported. Any eventually present alignment or correspondence
+#'   results will be dropped before combining the `XcmsExperiment` objects.
+#'   Finally, at present, only the MS data of the individual `XcmsExperiment`
+#'   objects is combined and any data eventually present in the `@qdata`,
```
sneumann, line 57: I haven't dug into the function, but what prevents `c()`ing also slots like `@experimentFiles` from two or more `XcmsExperiment`s ?

R/functions-XCMSnExp.R
sneumann: ok, convinced. The paranoid among us could check `colnames()[3]=="intensity" once before the loop.

sneumann (approved):
Hi, thanks everyone for the PR and review work. Yours, Steffen

R/XcmsExperiment-functions.R
jorainer: yes, indeed, true. was not paying attention actually. I tend to write internal function names now always in snake_case - for no particular reason except that I find it a bit more readable

R/XcmsExperiment-functions.R
jorainer: again, yes. I was using `.mse_` as an abbreviation for `MsExperiment` in internal function names - to quickly see from the function name that this is supposed to work on `MsExperiment` objects - then for `XcmsExperiment` I used `xmse` - and true, I sometimes even misspell it now. Would just need changes in a lot of places if we want to revert :(

R/XcmsExperiment.R
jorainer: Honestly - I never used the `@experimentFiles` slot - and am actually also not quite sure it will be used a lot...

PR #773, opened 2024-10-02
addition of chromPeaksId to `matchLamaChromPeaks()` output

philouail:
Here is a small change to the `matchLamaChromPeaks()` and its helper function which now means that the output of the match between the anchor peaks and the chromPeaks has an extra column called chromPeaksId. 

First time I make a small update like this, should I still bump the version ? Or you want to do that and mention this change in a next big update ? 

PS: The number of commits in this PR is due to a previous project I was working on, on this branch. It's been removed.


jorainer (approved):
Looks good to me! thanks!

jorainer (approved):
Thanks @philouail ! Looks good to me!

sneumann:
Looks good to me, too. Whichever PR gets merged first, the other needs to use a version `>4.3.4` :-) Yours, Steffen

jorainer:
OK, then I'll merge this into devel and update my branch/PR with this and change the version there.

PR #779, opened 2024-11-20
Adding `chromPeaks` metadata to the `Spectra` output of `chromPeakSpectra()`

philouail:
@jorainer Just wanted to get your opinion on this.. Because appending multiple column at once to the `spectraData()` of an object is not possible. i only see 2 options: 
- accessing and manipulating the spectraData slot ( in this PR)
- a  `for` loop adding columns one by ones to the `res` object. 

Am I missing an easier way to deal with this ? 

This PR is not ready, just wanted to get your opinion before I also update `featureSpectra()` and the documentation.

### DIFF
--- a/R/AllGenerics.R
+++ b/R/AllGenerics.R
@@ -417,15 +417,13 @@ setGeneric("chromPeakData<-", function(object, value)
 #' spectra for one chromatographic peak (or a `Spectra` of length 0 if no
 #' spectrum was found for the respective chromatographic peak).
 #'
-#' Parameters `addColumnsChromPeaks` allow the user to add specific metadata
+#' Parameters `chromPeakColumns` allow the user to add specific metadata
 #' columns from the chromatographic peaks (`chromPeaks`) to the returned
 #' spectra object. This can be useful to retain information such as retention
 #' time (`rt`), m/z (`mz`). The columns will be named as they is written in the
-#' `chromPeaks` object with a prefix that is defined by the parameter
-#' `addColumnsChromPeaksPrefix`. The *peak ID* (i.e., the row name of the
-#' peak in the `chromPeaks` matrix) is always added to the spectra object as
-#' metadata column `paste0(addColumnsChromPeaksPrefix,id)`, by default it will
-#' be `"chrom_peak_id"`.
+#' `chromPeaks` object with the prefix `"chrom_peak_"`. The *peak ID*
+#' (i.e., the row name of the peak in the `chromPeaks` matrix) is always added
+#' to the spectra object as a metadata column named `"chrom_peak_id"`.
 #'
 #' See also the *LC-MS/MS data analysis* vignette for more details and examples.
 #'
@@ -460,15 +458,10 @@ setGeneric("chromPeakData<-", function(object, value)
 #' @param return.type `character(1)` defining the type of result object that
 #'     should be returned.
 #'
-#' @param addColumnsChromPeaks `character` vector with the names of the columns
+#' @param chromPeakColumns `character` vector with the names of the columns
 #'    from `chromPeaks` that should be added to the returned spectra object.
 #'    The columns will be named as they are written in the `chromPeaks` object
-#'    with a prefix that is defined by the parameter
-#'    `addColumnsChromPeaksPrefix`. Defaults to `c("mz", "rt")`.
-#'
-#' @param addColumnsChromPeaksPrefix `character(1)` defining the prefix that
-#'    should be used for the columns from `chromPeaks` that are added to the
-#'    returned spectra object. Defaults to `"chrom_peak_"`.
+#'    with a prefix `"chrom_peak_"`. Defaults to `c("mz", "rt")`.
 #'
 #' @param BPPARAM parallel processing setup. Defaults to [bpparam()].
 #'
@@ -517,7 +510,7 @@ setGeneric("chromPeakData<-", function(object, value)
 #' ms2_sps <- chromPeakSpectra(dda)
 #' ms2_sps
 #'
-#' ## spectra variable *peak_id* contain the row names of the peaks in the
+#' ## spectra variable *chrom_peak_id* contain the row names of the peaks in the
 #' ## chromPeak matrix and allow thus to map chromatographic peaks to the
 #' ## returned MS2 spectra
 #' ms2_sps$chrom_peak_id
@@ -817,23 +810,20 @@ setGeneric("featureDefinitions<-", function(object, value)
 #' spectra per feature).
 #'
 #' The information from `featureDefinitions` for each feature can be included
-#' in the returned [Spectra()] object using the `addColumnsFeatures` parameter.
+#' in the returned [Spectra()] object using the `featureColumns` parameter.
 #' This is useful for retaining details such as the median retention time (`rtmed`)
 #' or median m/z (`mzmed`). The columns will retain their names as specified
-#' in the `featureDefinitions` object, prefixed by the value of the
-#' `addColumnsFeaturesPrefix` parameter. Additionally, the *feature ID*
-#' (i.e., the row name of the feature in the `featureDefinitions` data.frame)
-#' is always added as a metadata column with the name
-#' `paste0(addColumnsFeaturesPrefix, "id")`, which defaults to `"feature_id"`.
+#' in the `featureDefinitions` object, prefixed by `"feature_"`
+#' (e.g., `"feature_mzmed"`). Additionally, the *feature ID* (i.e., the row
+#' name of the feature in the `featureDefinitions` data.frame) is always added
+#' as a metadata column named `"feature_id"`.
 #'
 #' See also [chromPeakSpectra()], as it supports a similar parameter for
 #' including columns from the chromatographic peaks in the returned spectra object.
 #' These parameters can be used in combination to include information from both
 #' the chromatographic peaks and the features in the returned [Spectra()].
 #' The *peak ID* (i.e., the row name of the peak in the `chromPeaks` matrix)
-#' is added as a metadata column with the name
-#' `paste0(addColumnsChromPeaksPrefix, "id")`, which defaults to
-#' `"chrom_peak_id"`.
+#' is added as a metadata column named `"chrom_peak_id"`.
 #'
 #' @param object [XcmsExperiment] or [XCMSnExp] object with feature defitions.
 #'
@@ -846,15 +836,11 @@ setGeneric("featureDefinitions<-", function(object, value)
 #'     `featureDefinitions(x)`). This parameter overrides `skipFilled` and is
 #'     only supported for `return.type` being either `"Spectra"` or `"List"`.
 #'
-#' @param addColumnsFeatures `character` vector with the names of the columns
+#' @param featureColumns `character` vector with the names of the columns
 #'     from `featureDefinitions` that should be added to the returned spectra
 #'     object. The columns will be named as they are written in the
-#'    `featureDefinitions` object with a prefix that is defined by the parameter
-#'    `addColumnsFeaturesPrefix`. Defaults to `c("mzmed", "rtmed")`.
-#'
-#' @param addColumnsFeaturesPrefix `character(1)` defining the prefix that
-#'     should be used for the columns from `featureDefinitions` that are added
-#'     to the returned spectra object. Defaults to `"feature_"`.
+#'    `featureDefinitions` object with the prefix `"feature_`.
+#'     Defaults to `c("mzmed", "rtmed")`.
 #'
 #' @param ... additional arguments to be passed along to [chromPeakSpectra()],
 #'     such as `method`.
@@ -866,7 +852,7 @@ setGeneric("featureDefinitions<-", function(object, value)
 #' the order and the length matches parameter `features` (or if no `features`
 #' is defined the order of the features in `featureDefinitions(object)`).
 #'
-#' Spectra variables `"peak_id"` and `"feature_id"` define to which
+#' Spectra variables `"chrom_peak_id"` and `"feature_id"` define to which
 #' chromatographic peak or feature each individual spectrum is associated
 #' with.
 #'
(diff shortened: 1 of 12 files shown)


R/XcmsExperiment-functions.R
```diff
@@ -830,6 +831,9 @@
             ids <- rep(rownames(pk), lengths(idx))
             res <- sp[unlist(idx)]
             res$peak_id <- ids
+            info <- pk[res$peak_id, peaksInfo]
+            colnames(info) <- paste("peak_", peaksInfo, sep = "")
+            res@backend@spectraData <- cbind(res@backend@spectraData, info)
```
jorainer: please don't use that - this will break for any `MsBackend` not extending `MsBackendDataFrame` or `MsBackendMemory`. The [`joinSpectraData()`](https://github.com/rformassspectrometry/Spectra/blob/main/R/Spectra-functions.R#L603-L660) function would be a possible way to merge them.

R/XcmsExperiment.R
```diff
@@ -1228,7 +1232,7 @@ setMethod(
     function(object, method = c("all", "closest_rt", "closest_mz",
                                 "largest_tic", "largest_bpi"),
              msLevel = 2L, expandRt = 0, expandMz = 0, ppm = 0,
-             skipFilled = FALSE, peaks = character(),
+             skipFilled = FALSE, peaks = character(), peaksInfo = c("rt", "mz"),
```
jorainer: I am not particularly happy with the name of this parameter `peaksInfo` - maybe something like `chromPeakColumns` or even `addColumns`/`addChromPeakColumns` would be more self-explanatory? Also, I'm a bit worried what might happen if we add a spectra variable `"mz"` to the `spectraData` - not sure if we don't get unexpected errors, because `mz` is considered to be a peaks variable, not a spectra variable, so it could be that some functionality breaks somewhere. Related to that, should we add a prefix like `"chrom_peak_"` to each of these column names? also maybe to distinguish them from feature columns, that might have the same name.

happy to discuss.

jorainer (changes requested):
this will be a very useful and helpful addition Phili! thanks! but there are some more general thing we can/should discuss

R/XcmsExperiment-functions.R
philouail: a damn it,  thanks I did not find this function yesterday. I expected accessing the slot would not be a good idea... Will update this thanks

R/XcmsExperiment.R
philouail: Regarding the columns name, It would indeed create issues with the already present  `mz`, I already have the helper function do something like that so the column names have `peaks_` in front of them. so it follow the previous pattern of `peaks_id`. Are you okay with that ? 

Yes i'm okay with changing the variable name, however, quick question. Because i will implement the same thing in `featureSpectra()`. Should it be the same variable ? in this case the name needs to be neutral (why i chose `peaksInfo`). So maybe `addColumns` ? and we make it clear that in the case of chrompeaks it will come from `chromPeaks()` and for feature it will come from `featureDefinitions()` ?

philouail: Because I don't know au I feel about have 2 different variables in the case of `featureSpectra()` (e.g. one such as `addfeatureColumns` and `addDhromPeaksColumns`) It just sound overly complicated and if someone ask for `featureSpectra()` I would expect they will use the feature informations no ?

jorainer: hm, yes, I see the point of having just one parameter `addColumns` for both... the documentation then has to clearly say what this refers to in the case of `featureSpectra()` and what for `chromPeakSpectra()`. But before settling on one common parameter, please rethink if we don't need to add any `chromPeak`-related columns if calling `featureSpectra()` (`featureSpectra()` is calling `chromPeakSpectra()`, so it would be easy to pass an (additional) parameter along).

For the prefix - I would suggest to have a parameter that would allow changing that but that has a meaningful prefix. So e.g. `addColumnsPrefix = "chrom_peak_"` for `chromPeakSpectra()` and `addColumnsPrefix = "feature_"` for `featureSpectra()`. I think it would be important to use self-explaining spectra variable names so that it's clear what is meant if one has only the `Spectra` object. E.g. for `sps$chrom_peak_rt` it would be pretty clear that the rt reported there is from the chrom peak.

philouail: I mean i don't really see anything against implementing adding columns from both chrompeaks and features definitions, just wanted to avoid having too many parameters/columns. but it could come in the future that someone needs both info so might as well implement it now, I will do that.

Regarding the prefix, is this something the user need to decide ?  again i'm a bit worried by the amount of parameters/ their length. featureSpectra would then have: 
addColumnsChrompeaks = , addColumnsChromPeaksPrefix =, addColumnsFeature = , addColumnsFeaturePrefix = ... 

I'll start writing the code for it and see how it looks

R/do_adjustRtime-functions.R
```diff
@@ -874,7 +874,7 @@ NULL
                       resid_ratio = 3,
                       zero_weight = 10,
                       bs = "tp"){
-    rt_map <- rt_map[order(rt_map$obs), ]
+    rt_map <- rt_map[order(rt_map$obs), c("ref", "obs")]
```
philouail, line 877: was this not added in a previous push ? @jorainer

philouail: same with the vignette update below ?

philouail:
@jorainer do you know what's happening with the Linux checks ? i get a weird error, did I do something wrong  ?

R/XcmsExperiment.R
jorainer: we could also hard-code the prefixes for now and have parameters later if needed.

R/do_adjustRtime-functions.R
jorainer: honestly - I also got a bit confused what we have now in the devel branch and what not...

jorainer: hm, maybe it's then in one of the other open pull requests?

philouail: yup that's exactly that ! I'll fix it if your PR gets merge first

R/AllGenerics.R
```diff
@@ -411,14 +411,21 @@ setGeneric("chromPeakData<-", function(object, value)
...
-#' element of the list contains thus a `Spectra` with all spectra for one
-#' chromatographic peak (or a `Spectra` of length 0 if no spectrum was found
-#' for the respective chromatographic peak).
+#' matching spectra is returned. With `return.type = "Spectra"` a `List` of
+#' `Spectra` is returned. The length of the list is equal to the number of rows
+#' of `chromPeaks`. Each element of the list contains thus a `Spectra` with all
+#' spectra for one chromatographic peak (or a `Spectra` of length 0 if no
+#' spectrum was found for the respective chromatographic peak).
+#'
+#' Parameters `addColumnsChromPeaks` allow the user to add specific metadata
```
jorainer: or should we call the parameter `chromPeakColumns`? Wouldn't `chromPeakSpectra(x, chromPeakColumns = c("rt", "rtmin"))` feel a bit more natural than `chromPeakSpectra(x, addColumnsChromPeaks = c("rt", "rtmin"))`?

but I don't have a strong opinion; happy to discuss.

R/AllGenerics.R
```diff
@@ -411,14 +411,21 @@ setGeneric("chromPeakData<-", function(object, value)
...
+#' columns from the chromatographic peaks (`chromPeaks`) to the returned
+#' spectra object. This can be useful to retain information such as retention
+#' time (`rt`), m/z (`mz`). The columns will be named as they is written in the
+#' `chromPeaks` object with a prefix that is defined by the parameter
+#' `addColumnsChromPeaksPrefix`. The *peak ID* (i.e., the row name of the
```
jorainer: hm. maybe we drop the `addColumnsChromPeaksPrefix` parameter for now and hard-code the prefix?

R/AllGenerics.R
```diff
@@ -503,7 +520,7 @@ setGeneric("chromPeakData<-", function(object, value)
 #' ## spectra variable *peak_id* contain the row names of the peaks in the
```
jorainer: note: you should also change here *peak_id* to *chrom_peak_id*

R/AllGenerics.R
```diff
@@ -799,10 +816,24 @@ setGeneric("featureDefinitions<-", function(object, value)
 #' spectrum **per chromatographic peak** will be returned (hence multiple
 #' spectra per feature).
 #'
-#' The ID of each chromatographic peak (i.e. its row name in `chromPeaks`)
-#' and each feature (i.e., its row name in `featureDefinitions`) are
-#' available in the returned [Spectra()] with spectra variables `"peak_id"`
-#' and `"feature_id"`, respectively.
+#' The information from `featureDefinitions` for each feature can be included
+#' in the returned [Spectra()] object using the `addColumnsFeatures` parameter.
```
jorainer: also here - maybe use `featureColumns` instead? or turn it round and use `addFeatureColumns`?

jorainer (commented):
Looks good to me, thanks Phili! Only thing is that I find the parameter names a bit bulky - `addColumnsChromPeaks` and `addColumnsFeatures` - I would prefer either `chromPeakColumn`/`featureColumns` or `addChromPeakColumns`/`addFeatureColumns`.

Also, I think you were right with the prefix. Maybe we don't expose that parameter in the function - I would still use it internally, so we can re-add it as a function parameter if needed, but for now I would remove it from the function definition.

R/AllGenerics.R
philouail: ah I knew i missed some ! I'll check around to be sure i didn't miss anymore

R/AllGenerics.R
philouail: I'm going with chromPeakColumns() it makes complete sense in my opinion

R/AllGenerics.R
philouail: yes it also weirdly complicates the code a bit aha. I'll remove that

R/AllGenerics.R
philouail: `featureColumns` is a bit less obvious than `chromPeakColumns` I think but the equivalent would be `featureDefinitionsColumns` and that is way too long. let's go with `featureColumns`, the description is clear enough, user will understand.

R/AllGenerics.R
philouail: ah shoot i just re-read you message above. and you didn't want me to remove it completely... I mean it maintain that it over-complicates the code. So I would actually remove it. i'll be happy to change it back if needed later.

R/AllGenerics.R
```diff
@@ -411,14 +411,19 @@ setGeneric("chromPeakData<-", function(object, value)
...
-#' element of the list contains thus a `Spectra` with all spectra for one
-#' chromatographic peak (or a `Spectra` of length 0 if no spectrum was found
-#' for the respective chromatographic peak).
+#' matching spectra is returned. With `return.type = "Spectra"` a `List` of
+#' `Spectra` is returned. The length of the list is equal to the number of rows
+#' of `chromPeaks`. Each element of the list contains thus a `Spectra` with all
+#' spectra for one chromatographic peak (or a `Spectra` of length 0 if no
+#' spectrum was found for the respective chromatographic peak).
+#'
+#' Parameters `chromPeakColumns` allow the user to add specific metadata
```
jorainer: *Parameter `chromPeakColumns` allows*

R/AllGenerics.R
```diff
@@ -411,14 +411,19 @@ setGeneric("chromPeakData<-", function(object, value)
...
+#' columns from the chromatographic peaks (`chromPeaks`) to the returned
+#' spectra object. This can be useful to retain information such as retention
```
jorainer: Maybe use here *keep* instead of *retain* to avoid using *retain* next to *retention time*

R/AllGenerics.R
```diff
@@ -411,14 +411,19 @@ setGeneric("chromPeakData<-", function(object, value)
...
+#' time (`rt`), m/z (`mz`). The columns will be named as they is written in the
```
jorainer: *as they are written*

R/XcmsExperiment-functions.R
```diff
@@ -793,7 +793,9 @@
                                                  "largest_bpi"),
                                    msLevel = 2L, expandRt = 0, expandMz = 0,
                                    ppm = 0, skipFilled = FALSE,
-                                   peaks = integer(), BPPARAM = bpparam()) {
+                                   peaks = integer(),
+                                   chromPeakColumns = c("rt", "mz"),
+                                   BPPARAM = bpparam()) {
     method <- match.arg(method)
     pks <- .chromPeaks(x)[, c("mz", "mzmin", "mzmax", "rt",
```
jorainer: you need to add a check whether there are all `chromPeakColumns %in% colnames(.chromPeaks(x))`. Also, then you should ensure that these columns are also returned in following call, i.e. `pks <- .chromPeaks(x)[, union(c("mz", "mzmin", "mzmax", "rt", "rtmin", "rtmax", "maxo", "sample"), chromPeakColumns)]`

R/XcmsExperiment.R
```diff
@@ -1773,7 +1775,9 @@ setMethod(
     "featureSpectra", "XcmsExperiment",
     function(object, msLevel = 2L, expandRt = 0, expandMz = 0, ppm = 0,
              skipFilled = FALSE, return.type = c("Spectra", "List"),
-             features = character(), ...) {
+             features = character(),
+             featureColumns = c("rtmed", "mzmed"),
+             ...) {
         return.type <- match.arg(return.type)
         if (!hasFeatures(object))
```
jorainer, line 1782: I would also suggest here to first check that `featureColumns` are indeed available in `featureDefinitions()`

jorainer (changes requested):
Thanks for adapting, but now I found some additional issues that might cause problems later, i.e. it would be key to check that the requested columns are available. sorry that I missed that the previous time.

jorainer (approved):
Thanks Phili! Looks great! Only very minor thing: please bump the version to 4.5.2.

@sneumann , would be OK for me to merge into *devel* and push to BioC

sneumann:
looks great, merging and pushing now. Yours, Steffen

PR #780, opened 2024-11-20
Fix/issue778 using Rf_error()

sneumann:


jorainer:
Thanks @sneumann for looking into this! I checked also locally using the current Bioconductor devel container, but there seem to be more C++ errors now - also in other places. I will check and push to this branch.

jorainer (approved):
Thanks Steffen! Locally (using the Bioconductor devel docker image) R CMD check completes without warnings or errors, so should be good to go.

PR #787, opened 2025-02-12
fix: manualChromPeaks to support integrating a single peak

jorainer:
This PR adds a fix to avoid an error when running `manualChromPeaks()` to integrate a single peak.

It includes a unit test, but @mar-garcia , please check that this fixes your error too.


sneumann (approved):
Yes, looks good to me.

Issue #789, opened 2025-03-12
Remove usage of deprecated and removed MSnbase::NAnnotatedDataFrame class


jorainer:
The `NAnnotatedDataFrame` class was now removed from the *MSnbase* package (in commit https://github.com/lgatto/MSnbase/commit/19426350bb6456b4cd35561d499184630317e3c8). We thus need to replace usage of the `NAnnotatedDataFrame` class with the `AnnotatedDataFrame` from `Biobase`.

jorainer:
Closing as it is fixed now.

PR #790, opened 2025-03-12
fix: replace NAnnotatedDataFrame with AnnotatedDataFrame

jorainer:
- This fixes issue #789 by replacing the use of the deprecated and removed `NAnnotatedDataFrame` with the `Biobase::AnnotatedDataFrame`.


R/XcmsExperiment-functions.R
```diff
@@ -1298,7 +1298,7 @@ XcmsExperiment <- function() {
                             processing = paste0("Data converted [", date(),"]"),
                             files = fileNames(from),
                             smoothed = NA)
-    n@phenoData <- new("NAnnotatedDataFrame", as.data.frame(sampleData(from)))
+    n@phenoData <- AnnotatedDataFrame(as.data.frame(sampleData(from)))
```
sneumann, line 1301: using `new()` vs. directly the constructor, is that a style thing, or are these different class systems ?

R/methods-XCMSnExp.R
```diff
@@ -2404,9 +2404,9 @@ setMethod("fillChromPeaks",
...
                       processingData = new("MSnProcess",
                                            files = fileNames(object)[i]),
-                      featureData = new("AnnotatedDataFrame", fd),
-                      phenoData = new("NAnnotatedDataFrame",
-                                      data.frame(sampleNames = "1")),
+                      featureData = AnnotatedDataFrame(fd),
+                      phenoData = AnnotatedDataFrame(
+                          data.frame(sampleNames = "1")),
                       experimentData = new("MIAPE",
                                            instrumentManufacturer = "a",
```
sneumann, line 2411: "1" and "a" are dummy values. Should one make that more visible ? "Dummy" ?

sneumann (approved):
Looks good to me

sneumann:
... and one more: the changes are implicitly tested in the tests on `XcmsExperiment` and `XCMSnExp` , right ? Yours, Steffen

R/methods-XCMSnExp.R
jorainer: AFAIK this is just some internal code to more efficiently process the data - the object with these values is never exported to the user.

jorainer:
The changes are not explicitly tested - but the `NAnnotatedDataFrame` class does no longer exist and MSnbase uses now the `AnnotatedDataFrame` class instead. Any code on an `XCMSnExp` was failing without this update. The `XcmsExperiment` is not affected by these changes.

R/XcmsExperiment-functions.R
jorainer: Creating new S4 objects using a constructor function is a more elegant, and safer way than using `new()`. It's more user friendly and allows also to check and eventually fix/format the input parameters. For the `MSnbase::NAnnotatedDataFrame` there was never a constructor function - in contrast to the `Biobase::AnnotatedDataFrame()`.

PR #794, opened 2025-04-28
Add support for (very) large-scale data

jorainer:
This PR adds a new *xcms* result object `XcmsExperimentHdf5` that stores all preprocessing results to a local file in HDF5 format. Thus, no data/results are kept in memory reducing the memory demand to a minimum.

For an example and performance evaluation of this backend check the *large_scale* branch on Metabonaut: https://github.com/rformassspectrometry/Metabonaut/tree/large_scale_data

There are also other, smaller, changes in this PR including:

- improve performance of `chromatogram` call by reading only spectra within the retention time ranges
- linking to Metabonaut
- update **all** documentations to roxygen/markdown. Fix links in documentation.

Apologies for the very large PR - there are unit tests for all functions, so the code should be OK. So maybe just discuss/evaluate the concept?


NAMESPACE
```diff
@@ -291,156 +294,6 @@ exportClasses(
     "FilterIntensityParam",
     "ChromPeakAreaParam"
 )
-## Param methods
```
sneumann, line 294: Were those removed entirely ? I see a few new exports elsewhere, but only a minority of these here

NEWS.md
```diff
@@ -1,11 +1,29 @@
+# xcms 4.7
+
+## Changes in version 4.7.1
+
+- Change the naming convention of chromatographic peaks to include also the MS
```
sneumann, line 5: That starts to put semantics into the identifier. I am extra careful with that, code needs to check MS level <=9. Are there any other semantics that could creep around the corner and need consideration ? Ion Mobility ? Probably not, remains an attribute of the peaks. Can you still rely on the leading digit in older datasets ? CP2001 could also be the 2001st peak in an old dataset ...

NEWS.md
```diff
@@ -1,11 +1,29 @@
...
+  level after `"CP"`: `"CP1001"` instead of `"CP001"` for a chromatographic peak
+  in MS level 1  or `"CP2001"` instead of `"CP001"` for a chromatographic peak
+  in MS level 2.
+- Massively reduce main memory demand of *xcms* with the new on-disk storage
+  mode of the preprocessing results: the new `XcmsExperimentHdf5` result object
+  stores all preprocessing results in an HDF5 file.
+- Optimization and performance improvements for extraction of chromatographic
+  data. This includes using `MsCoreUtils::reduce()`.
+- Restructure and clean-up of documentation.
+- Don't export unnecessary get/set methods for `Param` classes.
```
sneumann, line 15: Ok. Methods are not neccessary or their export ? Could we have any old users of those functions ?

R/DataClasses.R
```diff
@@ -638,112 +627,90 @@ setClass("CentWaveParam",
...
-#' \code{\link{peaksWithMatchedFilter}} for functions to perform matchedFilter
+#' [peaksWithMatchedFilter()] for functions to perform matchedFilter
 #' peak detection in purely chromatographic data.
 #'
 #' @references
 #' Colin A. Smith, Elizabeth J. Want, Grace O'Maille, Ruben Abagyan and
 #' Gary Siuzdak. "XCMS: Processing Mass Spectrometry Data for Metabolite
 #' Profiling Using Nonlinear Peak Alignment, Matching, and Identification"
-#' \emph{Anal. Chem.} 2006, 78:779-787.
+#' *Anal. Chem.* 2006, 78:779-787.
```
sneumann, line 705: Future citations could include the DOI.

R/DataClasses.R
```diff
@@ -1204,73 +1139,51 @@ setClass("MSWParam",
...
 #' @examples
 #'
 #' ## Create a param object
-#' p <- CentWavePredIsoParam(maxCharge = 4)
-#' ## Change snthresh parameter
-#' snthresh(p) <- 25
+#' p <- CentWavePredIsoParam(maxCharge = 4, snthresh = 25)
 #' p
-#'
+NULL
```
sneumann, line 1187: I'd love to learn what these two NULL do, and what's the difference between 'em.

R/DataClasses.R
```diff
@@ -1660,185 +1575,164 @@ setClass("MsFeatureData", contains = c("environment"),
...
+#' - [overlappingFeatures()] identify features that are overlapping or close
+#'   in the m/z - rt space (see respective help page).
 #'
-#' \item \code{\link{overlappingFeatures}} identify features that are
-#' overlapping or close in the m/z - rt space (see respective help page).
+#' - [xcms::quantify()]: extract feature intensities and put them, along
+#'   with feature definitions and phenodata information, into a
+#'   [SummarizedExperiment::SummarizedExperiment()]. See help page for details.
 #'
-#' \item \code{\link{quantify}} extract feature intensities and put them, along
```
sneumann, line 1771: Is there some kind of deprecation documentation ? Or porting guide to xcms-4.7.x ?

R/DataClasses.R
```diff
@@ -2186,4 +2083,8 @@ setClass("FilterIntensityParam",
 setClass("BetaDistributionParam",
          contains = "Param"
          )
-        
\ No newline at end of file
+
+############################################################
+## Class unions
+setClassUnion("characterOrNULL", c("character", "NULL"))
```
sneumann, line 2093: Ah, I was wondering where they went :-)

R/MsExperiment-functions.R
```diff
@@ -63,7 +63,8 @@
 .mse_find_chrom_peaks_sample <- function(x, msLevel = 1L, param, ...) {
     x <- filterMsLevel(x, msLevel)
     pkd <- Spectra::peaksData(x, columns = c("mz", "intensity"),
-                              f = factor(), BPPARAM = SerialParam())
+                              f = factor(), return.type = "list",
```
sneumann, line 66: This inserted argument could confuse calling code that does not use named paramters, but relies on position. Is there a check whether `return.type` is a `SerialParam`to avoid silent stupid errors ?

R/XcmsExperiment-functions.R
```diff
@@ -1155,15 +1158,21 @@ featureArea <- function(object, mzmin = min, mzmax = max, rtmin = min,
         keep <- keep &
             chromPeakData(
                 object, return.type = "data.frame")$ms_level %in% msLevel
+    which(keep & .is_chrom_peaks_within_mz_rt(pks, rt, mz, ppm, type))
+}
+
+.is_chrom_peaks_within_mz_rt <- function(x, rt = numeric(), mz = numeric(),
```
sneumann: Bikeshedding: `.is_chrom_peak_within_mz_rt` singular ? Whatever, but consistent with names elsewhere.

R/XcmsExperiment-functions.R
```diff
@@ -1173,20 +1182,23 @@ featureArea <- function(object, mzmin = min, mzmax = max, rtmin = min,
         if (is.finite(mz[2L]))
             mz[2L] <- mz[2L] + mz[2L] * ppm / 1e6
         if (type == "any")
-            keep <- keep & pks[, "mzmin"] <= mz[2L] & pks[, "mzmax"] >= mz[1L]
+            keep <- keep & x[, "mzmin"] <= mz[2L] & x[, "mzmax"] >= mz[1L]
```
sneumann, line 1185: I have difficulties finding `x`, and what happened to pks

R/XcmsExperiment.R
```diff
@@ -870,11 +887,24 @@ setMethod(
...
-    function(object, param, msLevel = 1L, chunkSize = 2L, ...,
-             BPPARAM = bpparam()) {
+    function(object, param, msLevel = 1L, chunkSize = 2L,
+             hdf5File = character(), ..., BPPARAM = bpparam()) {
         if (length(msLevel) > 1)
             stop("Currently only peak detection in a single MS level is ",
                  "supported", call. = FALSE)
+        if (length(hdf5File)) {
+            if (file.exists(hdf5File))
+                stop("File ", hdf5File, " already exists. Replacing results ",
```
sneumann: If I have a vignette, does that mean to be on the safe side I need to `system("rm file.hdf5")` in the beginning ? Alternatively, there could be a `force.overwrite=TRUE` parameter somewhere

R/XcmsExperimentHdf5-functions.R
```diff
@@ -0,0 +1,1666 @@
...
+        Spectra()
+}
+
+################################################################################
+##
+##        FEATURES THINGS
+##
+################################################################################
+
+## ## WE MIGHT ACTUALLY NOT NEED THIS!
```
sneumann, line 841: Happy to learn why we may or may not need it

R/XcmsExperimentHdf5-functions.R
```diff
@@ -0,0 +1,1666 @@
+#' Properties of the HDF5 file used for on-disk storage of xcms results:
```
sneumann, line 1: Is there versioning info about the "xcms schema version" of the hdf5 in case we change it in the future, and can gracefully reject files in the old schema ? Or convert / migrate files ?

R/XcmsExperimentHdf5-functions.R
```diff
@@ -0,0 +1,1666 @@
...
+#' @noRd
+.h5_feature_values_ms_level <- function(ms_level, x, method, value, intensity,
+                                        filled = TRUE) {
+    cn <- .h5_chrom_peaks_colnames(x, ms_level)
+    col <- switch(method,
+                  sum = value,
+                  medret = c(value, "rt"),
+                  maxint = c(value, intensity))
+    if (!all(col %in% cn))
+        stop("Not all requested columns available. Please make sure 'value' ",
```
sneumann, line 955: Debugging could be simplified if we print the missing colnames. 
Why could they be missing ? Missing processing step ? Or new "xcms schema version" ?

R/XcmsExperimentHdf5-functions.R
```diff
@@ -0,0 +1,1666 @@
...
+    msl[has_cp]
+}
+
+#' Read selected datasets from a HDF5 file.
+#'
+#' @note
+#'
+#' Setting row and column names requires additional read steps and is hence
+#' considerably slower than *just* importing the data. Also, if possible,
+#' consider importing only the required column(s) using parameter `column`.
```
sneumann: Bikeshedding again, could that be parameter `columns=c(...)` in plural ?

R/XcmsExperimentHdf5.R
```diff
@@ -0,0 +1,1501 @@
...
+#' @aliases XcmsExperimentHdf5-class
+#'
+#' @name XcmsExperimentHdf5
+#'
+#' @description
+#'
+#' The *xcms* result objects [XcmsExperiment()] and [XCMSnExp()] keep all
+#' preprocessing results in memory and can thus (depending on the size of the
+#' data set) require a large amount of memory. In contrast, the
+#' `XcmsExperimentHdf5` class, by using an on-disk data storage mechanism,
```
sneumann, line 14: How would the hdf5 approach actually work in, say, a Galaxy setting ? Where some of the operations can be executed in parallel, and then some operations (e..g correspondance analysis) join that into a single file.

R/XcmsExperimentHdf5.R
```diff
@@ -0,0 +1,1501 @@
...
+        xph <- XProcessHistory(
+            param = param, date. = date(), type. = .PROCSTEP.PEAK.REFINEMENT,
+            fileIndex = seq_along(object), msLevel = msLevel)
+        object <- addProcessHistory(object, xph)
+        object
+    })
+
+#' @rdname hidden_aliases
+setMethod("hasFilledChromPeaks", "XcmsExperimentHdf5", function(object) {
+    length(object@gap_peaks_ms_level) > 0
```
sneumann, line 665: So filled peaks are counted in gap_peaks. And what ms_level is that ? ms1 ?

R/XcmsExperimentHdf5.R
```diff
@@ -0,0 +1,1501 @@
...
+        object <- as(object, "MsExperiment") # need only a Spectra container
+        ## Need to iterate through files/samples
+        res <- lapply(seq_along(object), function(i) {
+            .h5_chrom_peak_spectra_sample(
+                h5f, ids[i], spectra(object[i]),
+                method = method, msLevel = msLevel, expandRt = expandRt,
+                expandMz = expandMz, ppm = ppm, skipFilled = skipFilled,
+                peaks = peaks, chromPeakColumns = chromPeakColumns)
+        })
+        res <- Spectra:::.concatenate_spectra(res)
```
sneumann, line 1293: Should that be exported by Spectra in the future ?

R/functions-XCMSnExp.R
```diff
@@ -577,35 +584,6 @@ dropGenericProcessHistory <- function(x, fun) {
     idxs
 }
 
-#' @rdname adjustRtime
-adjustRtimePeakGroups <- function(object, param = PeakGroupsParam(),
```
sneumann, line 581: This is entirely gone now ?

R/methods-xcmsFragments.R
```diff
@@ -3,6 +3,7 @@
 
 ############################################################
 ## show
+#' @rdname hidden_aliases
 setMethod("show", "xcmsFragments", .xcmsFragments.show)
```
sneumann, line 7: How useful is xcmsFragments nowadays ? Would this devel cycle be a good time to deprecate it and mark for removal in the next cycle ?

R/ramp.R
```diff
@@ -1,228 +0,0 @@
-## # rampInit <- function() {
```
sneumann, line 1: Rest in pieces :-)

tests/testthat/test_XcmsExperimentHdf5-functions.R
```diff
@@ -0,0 +1,1083 @@
+library(rhdf5)
```
sneumann, line 1: should the library() be in the parent directory, for all *hdf5 tests  ?

sneumann (commented):
Hi, awesome work. Some cosmetic questions here and there. Yours, Steffen

NAMESPACE
jorainer: I removed all of the *getter/setter* methods for the `*Param` classes (please let me know if there are any left - I would like to remove all).

Why I removed these:

- it clutters the documentation with seldomly, if at all, used functions
- users will almost never use these functions. It's *easier* to just create a new param object than changing the value of a previously created one.

NEWS.md
jorainer: The chrom peak identifiers were always just thought as an identifier **within** the same data set. So, IMHO, it does not matter what values these identifiers have (whether it's a running number or even just a random number, as long as it is unique). I agree that adding semantics to identifiers is not ideal - but for the new `XcmsExperimentHdf5` I needed something that allows me to define chromatographic peaks IDs separately for each process/file. Peak detection is performed on a per sample (file) basis and each of  these processes stores the identified peaks to the HDF5 file without knowing anything about the other projects. That's why I came up with this solution - to still have **unique** chrom peak IDs within the same data set. I'm of course open to any other working solution :)

NEWS.md
jorainer: Hard to tell if there are any users out there - I guess very few if any (we never really promoted the idea of using getter/setter methods of the parameter classes). I would change and see what happens - my guess is users come forward and will open issues, then we can address it.

R/DataClasses.R
jorainer: good point. I'll add the DOI for all references.

R/DataClasses.R
jorainer: The `NULL` after a roxygen documention just ensures that the documentation file (Rd) is created - without adding also the next function to it. For general documentation or collection/topics of function I use that, i.e., write the roxygen documentation and add a `NULL` at the end. Then, I *add* all related functions/methods/classes to the documentation using `#' @rdname <name of the general, topic, documentation>`

R/DataClasses.R
jorainer: I'm a bit confused what you mean. If this refers to the documentation, I did not delete any documentation entry, the `quantify()` method is described a bit above. I did not change/remove any old functionality. This is just some changes in the documentation, where I change to the newer *md*-type format - and eventually re-order functions alphabetically or group them by topic into *sections*.

R/DataClasses.R
jorainer: yes, wanted to somehow group them together.

R/MsExperiment-functions.R
jorainer: hm. good point. Or I move the new parameter `return.type` **after** `BPPARAM`?

R/XcmsExperiment-functions.R
jorainer: yes, agree - it sounds terrible :D . I'll change.

R/XcmsExperiment-functions.R
jorainer: `x` is the input parameter to `.is_chrom_peaks_within_mz_rt()`. It was `pks` before, but I renamed it to `x` because having the *main* input parameter to a function called `x` is more consistent with *general* R usage (same as the main parameter to a method is called `object`). I was not that consistent in the past, but for newer functions I try to stick to this more.

R/DataClasses.R
jorainer: I added the DOIs in commit https://github.com/sneumann/xcms/commit/c120f61f427c6cce120aa8a9e9225550cb98082b

R/XcmsExperiment-functions.R
jorainer: Done in https://github.com/sneumann/xcms/commit/c120f61f427c6cce120aa8a9e9225550cb98082b

R/XcmsExperiment.R
jorainer: Thanks! that's a great idea! I added a `force.overwrite` parameter in commit https://github.com/sneumann/xcms/commit/c120f61f427c6cce120aa8a9e9225550cb98082b

R/XcmsExperimentHdf5-functions.R
jorainer: This refers to the commented-out function below (`.h5_feature_chrom_peaks_sample()`). I implemented it but did not use in the end - I just wanted to keep the code as example/reference.

R/XcmsExperimentHdf5-functions.R
jorainer: I did not define a schema version - unless you can convince me otherwise, I think that will result in a bit of a large code overhead - also knowing that this HDF5 format is really just thought for very large experiments and will only be used sparsely.

R/XcmsExperimentHdf5-functions.R
jorainer: This is just fail safe - in case some (future?) other chrom peak detection algorithm is implemented that does not store these column names. I will add the supported column names.

R/XcmsExperimentHdf5-functions.R
jorainer: sure - I realized the parameter was actually called `name`, not `colname` - I will change to `names`.

jorainer: Indeed. I just cross-checked. The downstream functions use parameter `columns` not `column`, so this was just a mis-spelling in the description. I'll update.

R/XcmsExperimentHdf5-functions.R
jorainer: added in https://github.com/sneumann/xcms/commit/de9ae22946e7f5226ad2e5eaa4c30c6ddc4ca417

R/XcmsExperimentHdf5.R
jorainer: Parallel processing **must** be controlled by *xcms* (it's built in and can be configured with `BPPARAM` and `chunkSize` parameters). Data is processed in chunks of files, in parallel, results are then written, after processing one chunk, to the file. But that's similar to the *in-memory* approaches we have in *xcms*, where the result is collected at the end from the individual processes - only, here we don't do this at the very end of the processing (i.e. after processing all samples), but after processing the samples of one chunk.

R/XcmsExperimentHdf5.R
jorainer: no, the `@gap_peaks_ms_level` contains just the MS levels in which gap-filled peaks are present. So, if gap filling was performed on MS1 then we would have `@gap_peaks_ms_level = c(1L)`

R/XcmsExperimentHdf5.R
jorainer: I have to check the function - could be that it's already exported.

R/functions-XCMSnExp.R
jorainer: on the contrary - I converted that to a method - so I added dedicated implementations for `XcmsExperiment`, `XcmsExperimentHdf5` and `XCMSnExp`

R/methods-xcmsFragments.R
jorainer: I would be OK with that. I don't know what this class actually is/was used for... But we should maybe start deprecation in a separate branch/PR.

tests/testthat/test_XcmsExperimentHdf5-functions.R
jorainer: I wanted to keep the HDF5 functionality as separate as possible thus I did not add it into the main *testthat.R* file. I find it *cleaner* if the unit test file specifically sets the stage for its R source file.

R/MsExperiment-functions.R
jorainer: After checking the code again - this does not apply. This function is just called internally - without any possibility of the user to provide or change the parameters to this call.

R/XcmsExperimentHdf5.R
sneumann: This one without leading dot ? 
https://github.com/rformassspectrometry/Spectra/blob/82f773a3a8a305341b9baed3216632599649719d/NAMESPACE#L14

sneumann:
Excellent, thanks for the hard work and answering questions. Will merge next. Yours, Steffen

R/XcmsExperimentHdf5.R
jorainer: yes. I will update in the next PR.

Issue #796, opened 2025-05-15
bug in `chromPeakSpectra()` when `peaks =` parameters has duplicated input


philouail:
I will add a reproducible example asap. 

But essentialy as the title say, i have a bug with `chromPeakSpectra()` 

if the `return.type is = "Spectra"` and `method = "closest_rt" ` i get more spectra than in input chrompeaks. 
and if the `return.type = "List"`, the function fails.

philouail:
Here is the reproducible example: 

```r
> library(msdata)
>library(xcms)
> dda_file <- system.file("TripleTOF-SWATH", "PestMix1_DDA.mzML",
+                         package = "msdata")
> dda_data <- readMsExperiment(dda_file)
> cwp <- CentWaveParam(snthresh = 5, noise = 100, ppm = 10,
+                      peakwidth = c(3, 30))
> dda_data <- findChromPeaks(dda_data, param = cwp, msLevel = 1L)
[=======================================================================================================================================] 1/1 (100%) in 16s
> dda_spectra <- chromPeakSpectra(dda_data, msLevel = 1L, peaks = c("CP01", "CP02", "CP01"), method = "closest_rt")
> dda_spectra
MSn data (Spectra) with 5 spectra in a MsBackendMzR backend:
    msLevel     rtime scanIndex
  <integer> <numeric> <integer>
1         1   130.615      1037
2         1   130.615      1037
3         1   240.897      1843
4         1   130.615      1037
5         1   130.615      1037
 ... 37 more variables/columns.

file(s):
PestMix1_DDA.mzML
Processing:
 Filter: select MS level(s) 1 [Thu May 22 12:02:54 2025]
 Merge 1 Spectra into one [Thu May 22 12:02:54 2025] 
``` 

So i get more spectra than i input peaks...

philouail:
Oh and here is the error if i put return.type = "List"

> dda_spectra <- chromPeakSpectra(dda_data, msLevel = 1L, peaks = c("CP01", "CP02", "CP01"), method = "closest_rt", return.type = "List")
Error in h(simpleError(msg, call)) : 
  error in evaluating the argument 'f' in selecting a method for function 'split': factor level [3] is duplicated

jorainer:
Thanks! I'll look into this.

jorainer:
This was closed with PR #797 and #798

PR #797, opened 2025-05-26
fix issue in chromPeakSpectra for XcmsExperiment when duplicated chrom peaks are requested.

jorainer:
This PR fixes issue #796 in the 3.21 Bioconductor release (*RELEASE_3_21* branch). For description of the problem see issue #796

### DIFF
--- a/R/XcmsExperiment.R
+++ b/R/XcmsExperiment.R
@@ -1269,7 +1269,7 @@ setMethod(
             pkidx <- .i2index(peaks, rownames(.chromPeaks(object)), "peaks")
         else pkidx <- integer()
         res <- .mse_spectra_for_peaks(object, method, msLevel, expandRt,
-                                      expandMz, ppm, skipFilled, pkidx,
+                                      expandMz, ppm, skipFilled, unique(pkidx),
                                       chromPeakColumns,
                                       BPPARAM)
         if (!length(pkidx))
@@ -1278,7 +1278,7 @@ setMethod(
         if (return.type == "Spectra")
             res <- res[as.matrix(findMatches(peaks, res$chrom_peak_id))[, 2L]]
         else
-            as(split(res, factor(res$chrom_peak_id, levels = peaks)), "List")
+            as(split(res, factor(res$chrom_peak_id))[peaks], "List")
     })
 
 #' @rdname reconstructChromPeakSpectra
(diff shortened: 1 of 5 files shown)


tests/testthat/test_XcmsExperiment.R
```diff
@@ -1005,6 +1005,51 @@ test_that("chromPeakSpectra works", {
...
+                 c(CP1 = "CP1", CP2 = "CP2", CP1 = "CP1"))
+    expect_equal(vapply(d, function(z) z$rtime, NA_real_),
+                 c(CP1 = a[[1L]]$rtime, CP2 = a[[2L]]$rtime[1L],
+                   CP1 = a[[1L]]$rtime[1L]))
+
+    ## Getting all spectra.
+    a <- chromPeakSpectra(tmp, msLevel = 1L, peaks = c("CP1", "CP2"))
+    expect_true(all(a$chrom_peak_id %in% c("CP1", "CP2")))
+    b <- chromPeakSpectra(tmp, msLevel = 1L, peaks = c("CP1", "CP2", "CP1"))
+    expect_true(length(b) > length(a))
```
philouail, line 1050: nice, that's perfect !

philouail (approved):

R/XcmsExperiment.R
```diff
@@ -1269,7 +1269,7 @@ setMethod(
             pkidx <- .i2index(peaks, rownames(.chromPeaks(object)), "peaks")
         else pkidx <- integer()
         res <- .mse_spectra_for_peaks(object, method, msLevel, expandRt,
-                                      expandMz, ppm, skipFilled, pkidx,
+                                      expandMz, ppm, skipFilled, unique(pkidx),
```
sneumann, line 1272: Would anyone be surprised to get fewer (number of) peaks than requested ? Maybe a half-sentance in the manpage ?

sneumann (approved):
Great, thanks for fixing.

R/XcmsExperiment.R
jorainer: Actually, with this fix the number of peaks is identical to the requested number. The `unique()` here just ensures that we're not duplicating in two places - here we get the unique peaks and then later we duplicate and re-order them into what was requested from the user.

PR #798, opened 2025-05-26
Fix issue in chromPeakSpectra for XcmsExperiment when duplicated chrom peaks are requested

jorainer:
This PR fixes issue #796 . It contains the same fix as in PR #797 . For description of the problem see discussion in issue #796


sneumann:
Wonderful, thx.

Issue #801, opened 2025-05-27
Subsetting a XcmsExperiment object after alignment does not revert retention times of detected chrom peaks


jorainer:
Calling e.g. `xmse[4, keepAdjustedRtime = FALSE]` on an `XcmsExperiment` object after retention time alignment will correctly revert the retention times, but not revert the `"rt"`, `"rtmin"` and `"rtmax"` values of chromatographic peaks. The retention time information for chromatographic peaks will still be the adjusted retention times. Note that this does only affect the `XcmsExperiment`, but not the `XCMSnExp` result object.

This is a clear bug we need to fix ASAP in both the release and devel branch!

A reproducible example:

```r
library(xcms)
library(MsExperiment)
library(testthat)
fls <- c(system.file('cdf/KO/ko15.CDF', package = "faahKO"),
                    system.file('cdf/KO/ko16.CDF', package = "faahKO"),
                    system.file('cdf/KO/ko18.CDF', package = "faahKO"))
df <- data.frame(mzML_file = basename(fls),
                 dataOrigin = fls,
                 sample = c("ko15", "ko16", "ko18"))
mse <- readMsExperiment(spectraFiles = fls, sampleData = df)
cwp <- CentWaveParam(noise = 10000, snthresh = 40, prefilter = c(3, 10000))
xmse <- findChromPeaks(mse, param = cwp)
pdp <- PeakDensityParam(sampleGroups = rep(1, 3))
xmse <- groupChromPeaks(xmse, param = pdp, add = FALSE)
xmse <- adjustRtime(xmse, PeakGroupsParam(span = 0.4))

## Subsetting with/without keeping adjusted retention times
a <- xmse[3, keepAdjustedRtime = TRUE]
b <- tmp[3, keepAdjustedRtime = FALSE]
    
## retention times are expected to be different
expect_false(all(rtime(a) == rtime(b)))

## retention times of chrom peaks would also be expected to be different
## but they are not
expect_false(all(chromPeaks(a)[, "rt"] == chromPeaks(b)[, "rt"]))
Error: all(chromPeaks(a)[, "rt"] == chromPeaks(b)[, "rt"]) is not FALSE

`actual`:   TRUE 
`expected`: FALSE
```

jorainer:
[...]

PR #802, opened 2025-05-27
Ensure retention times of chromatographic peaks are reverted for [,XcmsExperiment with keepAdjustRtime = FALSE

jorainer:
This PR fixes the issue #801 , i.e., ensures that the retention times of chromatographic peaks are also, along with the spectra's retention times, reverted for subset of `XcmsExperiment` using the `[` method and using parameter `keepAdjustedRtime = FALSE`.

In addition, this PR also changes the default for `keepAdjustedRtime` from `FALSE` to `TRUE` to match the default value of the related `filterFile()` function.

I would then also make a PR to the release branch, as this is an important bug fix we should also have in the BioC 3.21 release version.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: xcms
-Version: 4.7.1
+Version: 4.7.2
 Title: LC-MS and GC-MS Data Analysis
 Description: Framework for processing and visualization of chromatographically
     separated and single-spectra mass spectral data. Imports from AIA/ANDI NetCDF,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # xcms 4.7
 
+## Changes in version 4.7.2
+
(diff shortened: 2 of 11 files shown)


jorainer:
@sneumann , this PR contains now also a fix for issue #803 . In essence, this was mostly a unit test failure. With the recent update in the *MsExperiment* package the `sampleData()` `data.frame` gets automatically row names (the file names). For `chromPeakChromatograms()` sample names don't make sense (since one EIC is returned for each chrom peak, but there is no organization of these EICs by sample). The fix thus drops the sample names for `chromPeakChromatograms()`.

jorainer:
I added the same fix in the release version (see PR https://github.com/sneumann/xcms/pull/805)

sneumann:
Thanks, looking good. Yours, Steffen

jorainer:
Thanks for looking into this @sneumann ! - do you think we could merge and push to BioC?

Issue #804, opened 2025-06-09
Build error on release on macOS, check error on Linux and Windows


sneumann:
Hi,
on release, we have a build error on macOS both on `12.7.5 Monterey` and `13.6.6 Ventura` due to a missing header file, 
which might be an issue of the build machines. The error is absent on `lconway macOS 12.7.1 Monterey` and `kjohnson3 macOS 13.7.1 Ventura `

Also, we have a testthat error, which is likely the same as in devel #803.

Yours, Steffen

https://bioconductor.org/checkResults/release/bioc-LATEST/xcms/merida1-install.html
```
massifquant/nmath.h:74:10: fatal error: 'libintl.h' file not found
#include <libintl.h>
         ^~~~~~~~~~~
1 error generated.
```

```
  ══ Failed tests ════════════════════════════════════════════════════════════════
  ── Failure ('test_XcmsExperiment.R:1427:5'): chromPeaksChromatograms,XcmsExperiment works ──
  ref@.Data not equal to res@.Data.
  Attributes: < Component "dimnames": Component 2: 1 string mismatch >
```

jorainer:
Note: I fixed this massifquant error already in the devel branch. I will make a PR with the respective fix also for the RELEASE_3_21 branch.

Issue #803, opened 2025-06-09
Check error on devel on all platforms ref@.Data not equal to res@.Data


sneumann:
Hi,
We have a testthat error on all platforms (also in release #804).
Yours, Steffen

https://bioconductor.org/checkResults/devel/bioc-LATEST/xcms/nebbiolo2-checksrc.html
```
  ══ Failed tests ════════════════════════════════════════════════════════════════
  ── Failure ('test_XcmsExperiment.R:1427:5'): chromPeaksChromatograms,XcmsExperiment works ──
  ref@.Data not equal to res@.Data.
  Attributes: < Component "dimnames": Component 2: 1 string mismatch >
```

in 
https://github.com/sneumann/xcms/blob/5f77f39b4ec93b3f6a90a40d46a5b8630bbac958/tests/testthat/test_XcmsExperiment.R#L1425

jorainer:
Thanks - I will look into this. Seems there are some row/colnames different. Could be related to some updates in other packages (*MsExperiment*?). I'll figure it out.

jorainer:
Note: I'm pretty sure it's related to the recent change in `readMsExperiment()` of `*MsExperiment* that sets the row names of the `sampleData()` (which was added to have a better support for conversion to `SummarizedExperiment`). In that case it's all just about the unit tests - where the *reference* does not have rownames while the actual `sampleData()`/`rowData()`/`pData()` does.

PR #805, opened 2025-06-10
fix: chromPeakSpectra,XcmsExperiment for updated MsExperiment

jorainer:
- Fix `chromPeakSpectra()` for `XcmsExperiment` objects for *MsExperiment* package versions >= 1.10.1: don't report column (sample) names for the resulting `XChromatograms` (issue #803, #804).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # xcms 4.6
 
+## Changes in version 4.6.2
+
+- Fix issue in `chromPeakSpectra()` for `XcmsExperiment` reporting also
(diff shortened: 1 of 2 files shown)


sneumann:
Hi, what I find weird is that R CMD check works with both the current version and also with your fix. 
which dependency do I have to update locally to trigger the same issue ? Yours, Steffen

jorainer:
It should be the *MsExperiment* package - for the BioC release you would need version >= 1.10.1 , for Bioc devel >= 1.11.1

what version do you have locally (`packageVersion("MsExperiment")`)?

PR #806, opened 2025-06-19
fix: disable gettext in massifquant

jorainer:
- Compilers on macOS don't provide gettext anymore (or ship a different version that does no longer include the libintl.h). This fix disables use of libintl.h and fixes issue #804.


jorainer:
unit tests on macOS fail because of the circular dependency between faahKO and xcms.

sneumann:
Looks good to me. Not checked, as I only have Linux

Issue #807, opened 2025-07-29
Convert XcmsExperiment to XCMSnExp object error


CLUES-Emory:
Hello,
We are trying to coerce an XcmsExperiment object to XCMSnExp so we can use the CPC package to evaluate peak quality using different LC methods. However, we seem to be running into an error when using the following function to convert our findChromPeaks object to XCMSnExp: `as(ChromPeak_Object, "XCMSnExp")`

The error is as follows (with traceback):
```
Error in validObject(n) : 
  invalid class “XCMSnExp” object: Required columns: ‘precursorScanNum’ not present in featureData!
4. stop(msg, ": ", errors, domain = NA)
3. validObject(n)
2. asMethod(object)
1. as(step_1_res, "XCMSnExp")
```
Thanks! Session info below.
```

R version 4.5.1 (2025-06-13)
Platform: aarch64-apple-darwin20
Running under: macOS Sonoma 14.7

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/lib/libRlapack.dylib;  LAPACK version 3.12.1

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: America/New_York
tzcode source: internal

attached base packages:
[1] stats4    stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] cpc_0.1.0           pander_0.6.6        signal_1.8-1        MSnbase_2.34.1      S4Vectors_0.46.0    Biobase_2.68.0      BiocGenerics_0.54.0
 [8] generics_0.1.4      mzR_2.42.0          Rcpp_1.1.0          readxl_1.4.5        dplyr_1.1.4         MsExperiment_1.10.1 ProtGenerics_1.40.0
[15] xcms_4.6.3          BiocParallel_1.42.1

loaded via a namespace (and not attached):
 [1] DBI_1.2.3                   remotes_2.5.0               rlang_1.1.6                 magrittr_2.0.3              clue_0.3-66                
 [6] MassSpecWavelet_1.74.0      matrixStats_1.5.0           compiler_4.5.1              vctrs_0.6.5                 reshape2_1.4.4             
[...]

PR #814, opened 2025-09-12
fix: conversion from XcmsExperiment to XCMSnExp

jorainer:
- Fix conversion from `XcmsExperiment` to `XCMSnExp` (issue #807).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: xcms
-Version: 4.6.3
+Version: 4.6.4
 Title: LC-MS and GC-MS Data Analysis
 Description: Framework for processing and visualization of chromatographically
     separated and single-spectra mass spectral data. Imports from AIA/ANDI NetCDF,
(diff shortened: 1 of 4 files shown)


sneumann:
Ah, so changing to proper getter function. 
I checked where else we have `@backend` access in the code and found
https://github.com/sneumann/xcms/blob/244737684bc33f5e6510add30f2729e3c6f6d840/R/functions-XCMSnExp.R#L1579
=> Should that be changed as well ? 
Yours,
Steffen

jorainer:
yep - I will change that - it's actually also odd why we call `cbind` there - would be much easier with `$feature_id <- ...`. I'll test that and push.

jorainer:
Done

jorainer:
Ready for merging and pushing to Bioconductor @sneumann ;)

sneumann:
Thanks everyone for the efforts. Pushed to github and bioc RELEASE_3_21. Yours, Steffen

Issue #756, opened 2024-06-27
ppm paramenter for ChromPeakAreaParam


stanstrup:
Hi,

I was looking into the newish ChromPeakAreaParam and trying to understand how wide windows it actually creates compared to the old way. It is definitely better than the old way that would go all the way to zero. But it still goes a lot lower than I think is likely optimal.
So to start with the conclusion I think it would still be advisable to have a  minimum ppm window like in the old function.

So first I calculated what the windows will look like for each feature. The peak picking ppm was 30. But we see that some windows would go as low as 7 ppm for some features. From the analysis below I would argue that that could be too low.

```r
peaks <- cbind(chromPeaks(xset_g_r_g_fill), chromPeakData(xset_g_r_g_fill)) %>% 
          as_tibble %>% 
          mutate(filled = if_else(is_filled, "filled", "not filled"))
```

```r
feature_def <- featureDefinitions(xset_g_r_g_fill) %>% 
                as_tibble(rownames="feature_id") %>% 
                rename(f_mzmed = "mzmed", f_mzmin = "mzmin", f_mzmax = "mzmax", f_rtmed = "rtmed", f_rtmin = "rtmin", f_rtmax = "rtmax") %>% 
                unnest(peakidx)

peaks_sum <- peaks %>% 
                mutate(peakidx = 1:n()) %>% 
                full_join(feature_def, by = "peakidx") %>% 
                group_by(feature_id, filled) %>% 
                summarise(mz = quantile(mz, probs = 0.25, names = FALSE),
                          mzmin = quantile(mzmin, probs = 0.25, names = FALSE),
                          mzmax = quantile(mzmax, probs = 0.25, names = FALSE),
                          .groups = "drop"
                          )

peaks_sum_notfilled <- peaks_sum %>% 
  filter(filled == "not filled") %>% 
  mutate(ppm_dev = 1E6*(mzmax-mzmin)/mz)
```

```r
ggplot(peaks_sum_notfilled, aes(x=ppm_dev)) +
  geom_histogram(position="identity", binwidth = 1,linewidth = NA) + 
  xlab("ppm") + ggtitle("Histogram of ppm window across matched peaks")+ 
  scale_x_continuous(breaks = seq(0,1000,10), lim = c(0,max(peaks_sum_notfilled$ppm_dev))) +
  theme_bw()
```

![image](https://github.com/sneumann/xcms/assets/4532682/15cb1afc-22f5-45ef-af01-8de7165c3712)

<br>
[...]

PR #815, opened 2025-09-22
Add support for skipFilled in featureSpectra for XcmsExperiment and two bug fixes

jorainer:
In this PR:

- Fix conversion from `XcmsExperiment` objects to `XCMSnExp` (issue #807).
- Add parameter `columns` to `chromPeakData()` to allow extraction of selected columns.
- Add support for parameter `skipFilled` to `featureSpectra()` for `XcmsExperiment` and `XcmsExperimentHdf5` to avoid extraction of spectra for gap-filled chromatographic peaks.
- Fix: add missing import of the `spectraData()` method from *Spectra*.


jorainer:
@sneumann , sorry to be a bit pushy here, but we should try to manage getting this PR into *xcms* before the next Bioconductor release. This contains some important bug fixes.

jorainer:
Note: added in the meantime also an additional parameter `minMzWidthPpm` for `featureArea()` and `ChromPeakAreaParam()` to address issue #756 .

R/DataClasses.R
```diff
@@ -1544,9 +1544,16 @@ setClass("ChromPeakAreaParam",
          slots = c(rtmin = "function",
                    rtmax = "function",
                    mzmin = "function",
-                   mzmax = "function"),
-         contains = "Param"
-         )
+                   mzmax = "function",
+                   minMzWidthPpm = "numeric"),
+         contains = "Param",,
```
sneumann: what is the intention behind `,,` here ?

R/XcmsExperiment.R
```diff
@@ -1898,18 +1917,24 @@ setMethod(
...
         sps <- .mse_spectra_for_peaks(
             object, msLevel = msLevel, expandRt = expandRt,
-            expandMz = expandMz, ppm = ppm, skipFilled = skipFilled,
-            peaks = unique(pindex), ...)
+            expandMz = expandMz, ppm = ppm, peaks = unique(pindex), ...)
+
         mtch <- as.matrix(
             findMatches(sps$chrom_peak_id,
-                        rownames(.chromPeaks(object))[pindex]))
+                        rownames(xcms:::.chromPeaks(object))[pindex]))
```
sneumann: ::: should not be needed here

R/XcmsExperiment.R
```diff
@@ -2165,3 +2192,21 @@ setMethod(
...
+## fa <- featureArea(ftms)
+## fd <- featureDefinitions(ftms)
+## b <- xcms:::.rect_overlap(fa[, "rtmin"], fa[, "rtmax"],
+##                           fa[, "mzmin"], fa[, "mzmax"])
+## a <- lapply(b, function(z) {
+##     c(mzdiff = diff(range(fd$mzmed[z])),
+##       rtdiff = diff(range(fd$rtmed[z])))
+## }) |> do.call(what = rbind)
+##
+## Thoughts on how to combine overlapping features:
```
sneumann, line 2209: so within a file overlapping features would result from non-optimal feature detection. 
Across samples overlapping features could come from non-optimal correspondence analysis.

sneumann (approved):
tiny cosmetic comments

R/DataClasses.R
jorainer: Oh, wow! does not make any sense, true. I'll remove. Thanks!

R/XcmsExperiment.R
jorainer: good catch - was a leftover from some tests. I'll remove.

R/XcmsExperiment.R
jorainer: yes - we've had now a dataset where we've seen some strange overlap in features. the idea of this function would be sort of quality check of the correspondence settings. but I'm not yet sure whether that's helpful. so, I just jotted down the idea to not forget to look more into it.

sneumann: I like the idea of having a corner of QC methods being calculated, with later the option to export those as mzQC values. but indeed, not just yet.

PR #819, opened 2025-10-30
replaced misleading explanation of applyAdjustedRtime

stanstrup:
I assume "*in parallel* to the adjusted" should have been "*in parallel* to the raw" . I replaced with a text similar to that in DataClasses.R.


jorainer (approved):
Thanks @stanstrup , I'm OK with the change. I leave it to @sneumann to merge.

thanks!

Issue #823, opened 2026-01-23
Suggest adding rtCenterFun and mzCenterFun options in groupChromPeaks / PeakDensityParam


DrRuiLi:
When using `xcms` to group peaks with `PeakDensityParam`, the resulting feature RT is currently calculated as the **median of the grouped peaks**. In datasets where:

* Low-intensity peaks are present in most samples, and
* High-intensity peaks are present in fewer samples,

the **median RT can be pulled away from the major (intense) peaks**, resulting in inaccurate feature RT assignments.

Example:

* 30 samples
* 20 small peaks around 910 s (low intensity, present in most samples)
* Major peaks around 870 s (high intensity, 1000× stronger, present in 10 samples)

Current behavior: grouped feature RT → ~903 s (median of all peaks), which **does not reflect the major peak RT**.

**Current workaround:**
For m/z in `CentWaveParam`, there is `mzCenterFun = "wMean"` to calculate an intensity-weighted m/z, but **no equivalent parameter exists for MZ or RT in `PeakDensityParam`**.

**Suggestion:**
Add **`rtCenterFun`** and optionally **`mzCenterFun`** parameters to `PeakDensityParam` / `groupChromPeaks` with options like:

* `"median"` (current default)
* `"mean"`
* `"wMean"` (weighted by intensity)

This would allow feature RT (and optionally m/z) to better reflect **major/intense peaks**, improving grouping accuracy, especially in heterogeneous datasets.

**Benefits:**

* Prevents feature RT from being skewed by numerous low-intensity peaks
* Makes grouped features more representative of biologically relevant peaks
* Consistent interface with `CentWaveParam`’s `mzCenterFun`

**Note on relationship to `bw`:**

The proposed `rtCenterFun` parameter is **independent of `bw`**. While `bw` controls the density bandwidth and thus whether peaks are merged or split

* Increasing `bw` → broader grouping → risk merging distinct peaks
* Decreasing `bw` → narrower grouping → risk splitting the same peak across groups

 `rtCenterFun` would only determine **how the RT of an already grouped feature is calculated** (e.g., median, mean, intensity-weighted mean). This allows users to maintain their desired grouping behavior without being forced to adjust `bw` to compensate for RT skew caused by uneven peak intensities.

jorainer:
Thanks @DrRuiLi for this suggestion! we will add that parameter.

DrRuiLi:
Thanks a lot @jorainer for the quick response and implementation!
[...]

PR #828, opened 2026-02-23
Address issue 823

jorainer:
This PR:

- addresses issue #823 adding a parameter `rtCenterFun` to the `PeakDensityParam`.
- fix the title of the `plotPrecursorIons()` function.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: xcms
-Version: 4.9.2
+Version: 4.9.3
 Title: LC-MS and GC-MS Data Analysis
 Description: Framework for processing and visualization of chromatographically
     separated and single-spectra mass spectral data. Imports from AIA/ANDI NetCDF,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -63,7 +63,7 @@ Imports:
     MsCoreUtils (>= 1.19.2),
     MsFeatures,
     MsExperiment (>= 1.5.4),
-    Spectra (>= 1.16.1),
+    Spectra (>= 1.21.5),
     progress,
     RColorBrewer,
     MetaboCoreUtils (>= 1.11.2),
@@ -73,7 +73,6 @@ Suggests:
     caTools,
     knitr (>= 1.1.0),
     faahKO,
-    msdata (>= 0.25.1),
     ncdf4,
     testthat (>= 3.1.9),
     pander,
@@ -85,7 +84,8 @@ Suggests:
     MsBackendMgf,
     signal,
     mgcv,
-    rhdf5
+    rhdf5,
+    MsDataHub
 Enhances:
     Rgraphviz,
     rgl
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: xcms
-Version: 4.9.0
+Version: 4.9.1
 Title: LC-MS and GC-MS Data Analysis
 Description: Framework for processing and visualization of chromatographically
     separated and single-spectra mass spectral data. Imports from AIA/ANDI NetCDF,
--- a/R/XcmsExperiment.R
+++ b/R/XcmsExperiment.R
@@ -1584,8 +1584,10 @@ setMethod(
                  "using the 'findChromPeaks' method.", call. = FALSE)
         if (hasFeatures(object) && !add)
             object <- dropFeatureDefinitions(object)
-        cps <- chromPeaks(object, msLevel = msLevel,
-                          columns = c("mz", "rt", "sample"))
+        pk_cols <- c("mz", "rt", "sample")
+        if (.rtCenterFun(param) == "wMean")
+            pk_cols <- c(pk_cols, "into")
+        cps <- chromPeaks(object, msLevel = msLevel, columns = pk_cols)
         res <- .xmse_group_cpeaks(
             cps, param = param,
             index = match(rownames(cps), rownames(.chromPeaks(object))))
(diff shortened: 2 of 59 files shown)


jorainer:
Note @sneumann : this PR includes now also:

- remove dependency on the *msdata* R package: load all test files from *MsDataHub*. also the *longtests* were adapted accordingly.
- performance improvement for some functions by using `Spectra::rbindlistWithRownames()` to row-bind `data.frame`s.

longtests/testthat/test_methods-OnDiskMSnExp.R
```diff
@@ -102,7 +97,7 @@ test_that("findChromPeaks,OnDiskMSnExp,CentWavePredIsoParam works", {
 })
 
 test_that("findChromPeaks,OnDiskMSnExp,MassifquantParam works", {
-    mzf <- system.file("microtofq/MM14.mzML", package = "msdata")
+    mzf <- faahko_3_files[1L]
```
sneumann, line 100: Do the MsDataHub set of files have names instead of adressing by index ?

jorainer:
Wait - why are there now merge conflicts. I will have a look at them

R/do_groupChromPeaks-functions.R
```diff
@@ -107,7 +114,7 @@ do_groupChromPeaks_density <- function(peaks, sampleGroups,
         stop("'peaks' has to be a 'matrix' or a 'data.frame'!")
     ## Check that we've got all required columns
     .reqCols <- c("mz", "rt", "sample")
-    if (sleep > 0)
+    if (sleep > 0 | rtCenterFun == "wMean")
```
sneumann, line 117: Hm, sleep >0 was for displaying stuff during processing ? 
Ah, we need intensity either when displaying stuff or when calculating weights ? I was initially surprised why these quite different conditions were `or`ed.

R/XcmsExperiment.R
```diff
@@ -994,7 +995,7 @@ setMethod(
         lns <- lengths(pks)
         if (any(lns > 0)) {
             pks <- do.call(rbind, pks[lns > 0])
-            pkd <- do.call(rbind, lapply(res[lns > 0], function(z) {
+            pkd <- rbindlistWithRownames(lapply(res[lns > 0], function(z) {
```
sneumann, line 999: Is there a minimum dependency bump needed to use `rbindlistWithRownames` ?

tests/testthat.R
```diff
@@ -30,29 +35,25 @@ xod_chr <- findChromPeaks(filterMz(filterRt(od_x, rt = c(2500, 3500)),
                                    mz = c(334.9, 344.1)),
                           param = CentWaveParam())
 
-microtofq_fs <- c(system.file("microtofq/MM14.mzML", package = "msdata"),
-                  system.file("microtofq/MM8.mzML", package = "msdata"))
-microtofq_od <- readMSData(microtofq_fs, mode = "onDisk")
+## microtofq_fs <- c(system.file("microtofq/MM14.mzML", package = "msdata"),
```
sneumann: Do we need to get them into MsDataHub ?

DESCRIPTION
```diff
@@ -63,7 +63,7 @@ Imports:
     MsCoreUtils (>= 1.19.2),
     MsFeatures,
     MsExperiment (>= 1.5.4),
-    Spectra (>= 1.16.1),
+    Spectra (>= 1.21.5),
```
sneumann, line 66: Ah, found that Dependency bump

DESCRIPTION
```diff
@@ -85,7 +84,8 @@ Suggests:
     MsBackendMgf,
     signal,
     mgcv,
-    rhdf5
+    rhdf5,
+    MsDataHub
```
sneumann: Versioned MsDataHub ?

sneumann (approved):

R/XcmsExperiment.R
sneumann: Ok, found the version bump below

longtests/testthat/test_methods-OnDiskMSnExp.R
jorainer: yes, that's how I understood it. each file is one "data set", and they can be accessed by default through their name. But in this particular case I was defining the `faahko_3_files` variable in the main *testthat.R* file. It is 3 files from faahKO.

R/do_groupChromPeaks-functions.R
jorainer: yes exactly.

tests/testthat.R
jorainer: I did not want to add **all** files to MsDataHub - we can use other files for testing purposes too, the microtofq files were handy because they were small. There was no particular need or reason to use them (apart from the size). So I did not add them to MsDataHub but I changed to other test files. It's anyway better to use less different test files as they need to be downloaded and cached by the testing machine - so we save bandwidth.

DESCRIPTION
jorainer: good point. I will add that.

longtests/testthat/test_methods-OnDiskMSnExp.R
jorainer: Ah, now I got it. You wanted something like `faahko_3_files["ko15.CDF"]` instead of accessing by index. I think `faahko_3_files` does not have names, but I could directly use the `MsDataHub::ko15.CDF()` instead.

jorainer:
OK, @sneumann I fixed the merge conflicts. Also, I reverted to load the FAAH KO data files still from the *faahKO* R package. For now (this release) I think that's saver as we would need to change some additional test files, results files etc.

sneumann:
Cool, I can confirm it works locally. Merging and pushing now. 
Unsuer why we have a GH actions failure on Win+Mac, but that is not xcms related. 
YOurs, STeffen

PR #834, opened 2026-06-12
Avoid unnecessary/misleading warning in refineChromPeaks

jorainer:
`refineChromPeaks,XcmsExperiment,MergeNeighboringPeaksParam` was showing a warning about duplicated row names, but this was thrown by an internal merging of results - on temporary `data.frame`s/sub sets of the `chromPeakData()`. This PR is removing the row names from these intermediate objects. The final (real) row names of `chromPeakData()` were (and are) assigned correctly at the end of the call.


sneumann:
Thanks, both #834 and #835 have been merged and pushed to BioC. Yours, Steffen

PR #841, opened 2026-08-04
fix(filterFeatureDefinitions): remove all features on explicit empty selection

stanstrup:
# Fix: `filterFeatures()` keeps ALL features (instead of none) when the filter would remove every feature

## Summary

Every `filterFeatures()` parameter class computes the indices of the features to
**keep** and hands them to `filterFeatureDefinitions()`:

```r
fts_idx <- which(vals <= filter@threshold)          # features that PASS
...
filterFeatureDefinitions(object, features = fts_idx)
```

`filterFeatureDefinitions()` starts with a no-op guard:

```r
function(object, features = integer()) {
    if (!length(features))
        return(object)          # <-- treats "empty" as "no filter" => keep everything
    ...
}
```

So when a filter is strict enough that **no** feature passes (`fts_idx` is
`integer(0)`), `filterFeatureDefinitions()` returns the object **unchanged** —
i.e. it keeps **all** features instead of removing them. The user gets a
message claiming the features were removed, so the result silently contradicts
both the intent and the reported outcome.

This affects all three implementations (same guard):
`R/XcmsExperiment.R` (`XcmsExperiment`), `R/XcmsExperimentHdf5.R`
(`XcmsExperimentHdf5`), `R/methods-XCMSnExp.R` (`XCMSnExp`).

## Reproducible example

RSD is always `>= 0`, so `threshold = -1` passes **no** feature — every feature
should be removed:

```r
library(xcms)
library(MsExperiment)

x  <- loadXcmsData("xmse")                 # 351 features
qc <- which(sampleData(x)$sample_type == "QC")

r <- filterFeatures(x, RsdFilter(threshold = -1, qcIndex = qc))
#> 351 features were removed                <-- message says all removed
nrow(featureDefinitions(r))
#> [1] 351                                  <-- but ALL 351 are still there!
```

The filter works correctly whenever at least one feature passes; only the
"remove everything" case is wrong. The same applies to any direct call
`filterFeatureDefinitions(object, features = which(<condition matching nothing>))`.
With this fix the call above returns `0` features.

## Corroborating inconsistency

The sibling `filterChromPeaks()` handles the same "remove everything" case
**correctly**: it routes through `.filterChromPeaks()`, which does
`if (!length(idx)) return(new("MsFeatureData"))` — an explicit empty selection
empties the peak table. Likewise the **`SummarizedExperiment`** `filterFeatures`
methods subset with `object[fts_idx]`, so they handle the empty case correctly.
The identical call therefore gives **inconsistent results by object type**:

```r
filterFeatures(<SummarizedExperiment>, RsdFilter(threshold = -1, ...))  # -> 0 features (correct)
filterFeatures(<XcmsExperiment>,       RsdFilter(threshold = -1, ...))  # -> 351 features (BUG)
```

The fix distinguishes "no `features` argument given" (a genuine no-op) from "an
explicit empty selection" (keep none) via `missing(features)`; the no-argument
no-op is preserved.

### DIFF
--- a/tests/testthat/test_XcmsExperiment.R
+++ b/tests/testthat/test_XcmsExperiment.R
@@ -1243,10 +1243,12 @@ test_that("filterFeatureDefinitions works", {
     expect_true(hasFeatures(res))
     expect_equal(featureDefinitions(res), featureDefinitions(xmseg)[1:10, ])
 
-    ## An explicit empty selection removes ALL features - it must NOT be
-    ## treated as the no-argument no-op (which would silently keep them all).
     res0 <- filterFeatureDefinitions(xmseg, integer(0))
     expect_equal(nrow(featureDefinitions(res0)), 0L)
+    
+    res0B <- filterFeatures(xmseg, RsdFilter(threshold = -1, qcIndex = 1:2))
+    expect_equal(nrow(featureDefinitions(res0B)), 0L)
+    
 })
 
 test_that("featureSpectra works", {
--- a/tests/testthat/test_XcmsExperiment.R
+++ b/tests/testthat/test_XcmsExperiment.R
@@ -1242,6 +1242,11 @@ test_that("filterFeatureDefinitions works", {
     res <- filterFeatureDefinitions(xmseg, 1:10)
     expect_true(hasFeatures(res))
     expect_equal(featureDefinitions(res), featureDefinitions(xmseg)[1:10, ])
+
+    ## An explicit empty selection removes ALL features - it must NOT be
+    ## treated as the no-argument no-op (which would silently keep them all).
+    res0 <- filterFeatureDefinitions(xmseg, integer(0))
+    expect_equal(nrow(featureDefinitions(res0)), 0L)
 })
 
 test_that("featureSpectra works", {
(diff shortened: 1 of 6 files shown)


tests/testthat/test_XcmsExperiment.R
```diff
@@ -1242,6 +1242,11 @@ test_that("filterFeatureDefinitions works", {
     res <- filterFeatureDefinitions(xmseg, 1:10)
     expect_true(hasFeatures(res))
     expect_equal(featureDefinitions(res), featureDefinitions(xmseg)[1:10, ])
+
+    ## An explicit empty selection removes ALL features - it must NOT be
```
philouail: I would remove the comments in tests to follow the current format. 

also maybe add a test similar to the "reproducible example" you used in the PR text. So the method deployment that showed you the problem is also tested. 

```
r <- filterFeatures(x, RsdFilter(threshold = -1, qcIndex = qc))
expect_equal(..)
```

philouail: a bit superfluous but just to be sure :)

philouail (changes requested):
Can you update the test ?

stanstrup:
Thanks. Removed the comment, added a test going through filterFeatures as the example and, updated the version number and added a NEWS entry as suggested by @jorainer in one of the other PRs.

Issue #825, opened 2026-02-06
refineChromPeaks,MergeNeighboringPeaksParam fails to merge nested peaks


jorainer:
With `MergeNeighboringPeaksParam`, `refineChromPeaks()` should merge partially or completely overlapping peaks. In the current implementation it can however fail to merge the latter, depending on the order of the peaks in `chromPeaks()`:

```r
## Define overlapping/nested chrom peaks 
pks <- rbind(
    c(666.0693, 666.0693, 666.0693, 31.779, 27.59, 35.968, 1, 1, 1, 1, 1),
    c(666.0713, 666.0683, 666.0747, 31.181, 27.59, 39.559, 2, 2, 2, 2, 1),
    c(666.0693, 666.0693, 666.0693, 31.779, 27.59, 36.968, 3, 3, 3, 3, 1))
colnames(pks) <- c("mz", "mzmin", "mzmax", "rt", "rtmin", "rtmax", "into",
                   "intb", "maxo", "sn", "sample")
rownames(pks) <- c("A", "B", "C")
```

In this example, the first (`"A"`) and third (`"C"`) peak are completely within the *m/z* and rt range of the second (`"B"`). Peak merging should therefore only report the second, but not the first or third. The internal function to merge peaks is `xcms:::.merge_neighboring_peak_candidates()`.

```r
## define the remaining data required for the function
pkd <- data.frame(ms_level = rep(1L, 3), is_filled = rep(FALSE, 3))

x <- list(cbind(mz = c(), intensity = c()),
          cbind(mz = c(), intensity = c()),
          cbind(mz = c(), intensity = c()))
rt <- c(30.5, 31.5, 32.5)
```

Running this function with these data results in

```r
> xcms:::.merge_neighboring_peak_candidates(x, rt, pks, pkd)
$chromPeaks
        mz    mzmin    mzmax     rt rtmin  rtmax into intb maxo sn sample
A 666.0693 666.0693 666.0693 31.779 27.59 35.968    1    1    1  1      1
B 666.0713 666.0683 666.0747 31.181 27.59 39.559    2    2    2  2      1

$chromPeakData
  ms_level is_filled
1        1     FALSE
2        1     FALSE
```

so, both `"A"` and `"B"` are reported, although `"A"` is completely within `"B"`.

We get the expected results if we change the order of the peaks at input into B, A, C:

```r
> xcms:::.merge_neighboring_peak_candidates(x, rt, pks[c(2, 1, 3), ], pkd)
[...]

Issue #599, opened 2022-02-10
Why the findChromPeaks function does not recognize smaller peak?


Violet99999:
We are trying to use xcms to select peaks for analytes based on the pre-set retention time. In the plot below, the correct peak should be the lowest one at 4.15 minute; however, with our code, xcms always selects the highest peak. It seems that the 'findChromPeaks' function ignores the peak at 4.15 minutes for some reasons, even the intensity at 4.15 minute is not. We are wondering why the 'findChromPeaks' function doesn't recognize the smaller peak as a peak?  

Here is our code to detect possible peaks for analytes: 
` srmdata <- readSRMData(files=sample_file) # sample file is a mzml file exported directly from lcms machine`
`mz <- srmdata[i] # mz is one single chromatogram from the srmdata (large Mchromatogram)`
`matched_filter_param <- MatchedFilterParam(fwhm=0.001, snthresh=0)`
`xchr1 <- findChromPeaks(mz, param=matched_filter_param)`

xchr1 result:
`Object of class: XChromatogram`
`length of object: 83`
`from file: 1`
`mz range: [NA, NA]  #mz ratio is 225.1/44.2` 
`rt range: [3.636217, 4.644417]`
`MS level: 1`
`Identified chromatographic peaks (5):`
 `rt	rtmin	rtmax	into	maxo	sn (2 more column(s))`
` 4.39253333333333	3.9692	4.64441666666667	12082.268825	284397	15.6017699285388`
 `3.92975	3.90295	3.94316666666667	5.75098333333329	322	0.0176646375207526`
` 3.83626666666667	3.77858333333333	3.84998333333333	6.11184000000001	214	0.0117398522653459`
` 3.70573333333333	3.70573333333333	3.70573333333333	NaN	214	0.011739852265344`
` 3.8766	3.8766	3.8766	NaN	107	0.00586992613267289`

Set retention time as x and intensity as y 
`{"x":4.094016666666667,"y":0.0},`
`{"x":4.106583333333333,"y":107.0},`
`{"x":4.118783333333333,"y":2466.0},`
`{"x":4.130983333333333,"y":10509.0},`
`{"x":4.1432,"y":19839.0},`
**`{"x":4.1554,"y":22627.0},`** This is the top point of the correct peak but 'findChromPeaks' does not recognize it as a peak (see xchr1 result)
`{"x":4.167883333333333,"y":16836.0},`
`{"x":4.18035,"y":8686.0},`
[...]

PR #842, opened 2026-08-04
fix(refineChromPeaks): avoid into=Inf when merged peak spans <=1 scan

stanstrup:
# Fix: merging neighboring chromatographic peaks yields `into = Inf` when the merged peak spans ≤ 1 scan

## Summary

`refineChromPeaks(..., MergeNeighboringPeaksParam(...))` re-integrates a merged
peak as `sum(intensities) * peak_width`, where

```r
peak_width <- (rtmax - rtmin) / (idx_max - idx_min)
```

`idx_min`/`idx_max` are the indices of the retention times closest to the
merged peak's `rtmin`/`rtmax`. When the merged peak's retention-time window
covers only a **single** data point (`idx_max == idx_min`), this divides by
zero, so `peak_width` becomes `Inf` and the peak's `"into"` becomes **`Inf`**.
That `Inf` then silently propagates into `featureValues()`.

The same defect exists in both merge implementations:

- `R/functions-Chromatogram.R`, `.chrom_merge_neighboring_peaks()` (used for
  `XChromatogram` / `XChromatograms`): `… / (idx_max - idx_min)`.
- `R/XcmsExperiment-functions.R`, `.merge_neighboring_peak_candidates()` (used
  for `XcmsExperiment`): `peak_width <- (rtmax - rtmin) / (idx[length(idx)] - idx[1L])`.

This is triggered by sparse / coarse-in-rt chromatographic data (e.g.
flow-injection / direct-infusion data, fast low-scan-rate acquisitions, sparse
EICs after m/z filtering, or manually defined narrow peaks), where a merged
peak can span fewer than two scans.

## Related GitHub issue

- **Related to #825** ("refineChromPeaks,MergeNeighboringPeaksParam fails to
  merge nested peaks", open). That issue is a **different defect in the same
  function** (`.merge_neighboring_peak_candidates()`): its `order(pks[, "rtmin"])`
  sort leaves fully-nested peaks in input order so the smaller nested peak is not
  merged away. It is *not* the divide-by-zero fixed here.

## Reproducible example

```r
library(xcms)
library(MSnbase)

## A chromatogram with a COARSE rt grid (3 scans, 10 s apart) and two peaks
## whose merged rt window [18, 22] contains only ONE scan (rt = 20):
chr <- XChromatogram(rtime = c(10, 20, 30), intensity = c(100, 900, 100))
reqn <- c("mz", "mzmin", "mzmax", "rt", "rtmin", "rtmax", "into", "maxo", "sn", "sample")
pks  <- matrix(0, nrow = 2, ncol = length(reqn), dimnames = list(NULL, reqn))
pks[1, ] <- c(100, 100, 100, 20, 18, 21, 900, 900, 10, 1)
pks[2, ] <- c(100, 100, 100, 21, 20, 22, 500, 500, 10, 1)
chromPeaks(chr) <- pks
xchr <- XChromatograms(list(chr), nrow = 1, ncol = 1)

chromPeaks(xchr)[, "into"]
#> [1] 900 500

ref <- refineChromPeaks(xchr, MergeNeighboringPeaksParam(expandRt = 4, minProp = 0.01))
chromPeaks(ref)[, "into"]
#> CPM1
#>  Inf          <-- corrupt
```

With this fix the merged peak has a finite `into` of `3600`
(`sum(intensity in [18, 22]) = 900`, `peak_width = (22 - 18) / 1 = 4`).

### DIFF
--- a/R/XcmsExperiment-functions.R
+++ b/R/XcmsExperiment-functions.R
@@ -287,7 +287,7 @@
                 rtmin <- pks_new[current_peak, "rtmin"]
                 rtmax <- max(pks[i, "rtmax"], pks_new[current_peak, "rtmax"])
                 idx <- which(rt >= rtmin & rt <= rtmax)
-                peak_width <- (rtmax - rtmin) / (idx[length(idx)] - idx[1L])
+                peak_width <- (rtmax - rtmin) / max(1L, idx[length(idx)] - idx[1L])
                 pkm <- do.call(rbind, x[idx])
                 pks_new[current_peak, c("rtmax", "mzmin", "mzmax", "into")] <-
                     c(rtmax, range(pkm[, "mz"], na.rm = TRUE),
--- a/R/functions-Chromatogram.R
+++ b/R/functions-Chromatogram.R
@@ -127,7 +127,8 @@
                 idx_max <- which.min(
                     abs(rtime(x) - pks_new[current_peak, "rtmax"]))
                 peak_width <- (pks_new[current_peak, "rtmax"] -
-                               pks_new[current_peak, "rtmin"]) / (idx_max - idx_min)
+                               pks_new[current_peak, "rtmin"]) /
+                    max(1L, idx_max - idx_min)
                 ## Calculate into as done in centWave.
                 pks_new[current_peak, "into"] <-
                     sum(intensity(x)[rtime(x) >= pks_new[current_peak, "rtmin"] &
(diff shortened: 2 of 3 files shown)


philouail:
Quick note, i would not mention the issue in the PR notes as it seems this issue is NOT addressing that error no ?

philouail:
From what I understand this would mean that this would only be an issue through manual peak definitions ? 
I do not think any of the algorithm would return a peak that only span one scan (or even less zero). Or am I missing some other context ?

stanstrup:
Yes and No.

matchedFilter can produce peaks with rtmin == rtmax. centWave has a guard for the 0/0 that follows (if (is.na(pwid)) pwid <- 1), matchedFilter doesn't, so into comes out NaN. I suspect that's what's behind #599.

But after checking more, you are right. Even if you produce 1-scan peaks with matchedFilter they are not mishandled by MergeNeighboringPeaksParam.

PR #843, opened 2026-08-05
fix(fillChromPeaks): centWave gap-fill integrates over the actual scan span, not the requested window

stanstrup:
# Fix: centWave gap-filling integrates over the requested window instead of the actual scan span (inflates `into`)

## Summary

The centWave gap-fill integrators compute the filled peak's `"into"` as
`sum(intensities) × rt_width`, where `rt_width` is meant to be the mean per-scan
time step Δt in the Riemann sum `∫ I dt ≈ Δt · Σ I`. They compute Δt from the
**requested feature window**:

```r
rt_width <- (rtmax - rtmin) / (n_scans - 1)       # rtmax/rtmin = feature window
```

During gap filling the window `(rtmin, rtmax)` is the **cross-sample consensus**
feature window, which can be wider than the sample's actual scan coverage in
that window. `rt_width` is then larger than the true mean scan spacing, so
`into` is inflated by `requested_width / actual_span` and depends on how the
feature window happens to be padded.

This is inconsistent with two references that both use the **actual scan span**:

- **centWave detection** derives `into` from the peak's real scan times
  (`R/do_findChromPeaks-functions.R`):
  `pwid <- (scantime[hi] - scantime[lo]) / (hi - lo); into <- pwid * sum(pd)`,
  where `hi`/`lo` are the peak's actual boundary scans.
- The **matchedFilter gap-fill** already uses
  `diff(rt[idx_rt_range]) / diff(idx_rt_range)` — the actual data span.

Two centWave gap-fill functions are affected:

- `R/XcmsExperiment-functions.R`, `.chrom_peak_intensity_centWave()`
  (used by `fillChromPeaks()` on `XcmsExperiment`).
- `R/functions-XCMSnExp.R`, `.getChromPeakData()`
  (used by `fillChromPeaks()` on `XCMSnExp`).

The fix uses the actual rt span of the scans in the window, so gap-filled `into`
matches centWave detection and no longer depends on window padding. Values for
windows that already match the data span (the common case) are unchanged.

## Reproducible example

Integrating the **same four scans** with a tight window (`[10, 40]`, equal to
the data span) vs. a padded window (`[5, 45]`, 5 s wider on each side):

```r
library(xcms)

mkpm <- function(mz, i) cbind(mz = mz, intensity = i)
rt <- c(10, 20, 30, 40)
x  <- list(mkpm(c(100, 100.1), c(50, 60)), mkpm(c(100, 100.1), c(900, 800)),
           mkpm(c(100, 100.1), c(40, 30)), mkpm(c(100, 100.1), c(10, 5)))
cn <- c("mz", "mzmin", "mzmax", "rt", "rtmin", "rtmax", "into", "maxo", "sn", "sample")
pa <- function(rtmin, rtmax) matrix(
    c(100, 99.9, 100.2, (rtmin + rtmax) / 2, rtmin, rtmax), nrow = 1,
    dimnames = list("F", c("mz", "mzmin", "mzmax", "rt", "rtmin", "rtmax")))
into <- function(peakArea) xcms:::.chrom_peak_intensity_centWave(
    x, rt, peakArea, cn = cn, sampleIndex = 1L)[, "into"]

into(pa(10, 40))     # window == data span
into(pa(5, 45))      # same scans, wider window
```

Current source (before the fix):

```
tight [10,40] into = 18950
padded[5,45]  into = 25267      # 1.333x larger for the SAME signal (= 40/30)
```

With this fix both return `18950` — `into` no longer depends on the requested
window width, only on the signal actually present.

### DIFF
--- a/R/XcmsExperiment-functions.R
+++ b/R/XcmsExperiment-functions.R
@@ -557,7 +557,8 @@
                 res[i, c("rt", "mz", "maxo", "into")] <- c(
                     rts[maxi], mmz, mat[maxi, 2L],
                     sum(mat[, 2L], na.rm = TRUE) *
-                    (diff(range(rt[keep])) / max(1L, (length(keep) - 1L)))
+                    ((rt[keep[length(keep)]] - rt[keep[1L]]) /
+                     max(1L, (length(keep) - 1L)))
                 )
                 if ("beta_cor" %in% cn) {
                     res[i, c("beta_cor", "beta_snr")] <- .get_beta_values(
--- a/R/XcmsExperiment-functions.R
+++ b/R/XcmsExperiment-functions.R
@@ -557,7 +557,7 @@
                 res[i, c("rt", "mz", "maxo", "into")] <- c(
                     rts[maxi], mmz, mat[maxi, 2L],
                     sum(mat[, 2L], na.rm = TRUE) *
-                    ((rtr[2L] - rtr[1L]) / max(1L, (length(keep) - 1L)))
+                    (diff(range(rt[keep])) / max(1L, (length(keep) - 1L)))
                 )
                 if ("beta_cor" %in% cn) {
                     res[i, c("beta_cor", "beta_snr")] <- .get_beta_values(
--- a/R/functions-XCMSnExp.R
+++ b/R/functions-XCMSnExp.R
@@ -293,7 +293,7 @@ dropGenericProcessHistory <- function(x, fun) {
                 ## window overhangs the sample's local scan coverage. Numerator
                 ## and denominator use the same scan set (all scans in the
                 ## window); max(1, ...) guards the single-scan case.
-                sel <- which(rtim >= rtr[1] & rtim <= rtr[2])
+                sel <- which(between(rtim, rtr))
                 res[i, "into"] <- sum(mtx[, 3L], na.rm = TRUE) *
                     ((rtim[sel[length(sel)]] - rtim[sel[1L]]) /
                      max(1, (length(sel) - 1)))
--- a/R/functions-XCMSnExp.R
+++ b/R/functions-XCMSnExp.R
@@ -285,15 +285,18 @@ dropGenericProcessHistory <- function(x, fun) {
             if (any(!is.na(mtx[, 3]))) {
                 ## Area = sum(intensities) * rt_width, where rt_width is the mean
                 ## scan spacing = (actual rt span of the scans in the window) /
-                ## (number of those scans - 1). Using the *actual* data span
-                ## (rather than the requested window rtr[2] - rtr[1]) keeps the
-                ## filled 'into' consistent with centWave detection, which
-                ## derives pwid from the peak's actual scan times, and avoids
-                ## inflating 'into' when the feature window overhangs the
-                ## sample's local scan coverage. max(1, ...) guards single-scan.
-                rt_in <- rtim[rtim >= rtr[1] & rtim <= rtr[2]]
+                ## (number of those scans - 1). Using the actual span of the
+                ## scans in the window (rather than the requested window
+                ## rtr[2] - rtr[1]) keeps the filled 'into' consistent with
+                ## centWave detection, which derives pwid from the peak's actual
+                ## scan times, and avoids inflating 'into' when the feature
+                ## window overhangs the sample's local scan coverage. Numerator
+                ## and denominator use the same scan set (all scans in the
+                ## window); max(1, ...) guards the single-scan case.
+                sel <- which(rtim >= rtr[1] & rtim <= rtr[2])
                 res[i, "into"] <- sum(mtx[, 3L], na.rm = TRUE) *
-                    (diff(range(rt_in)) / max(1, (length(rt_in) - 1)))
+                    ((rtim[sel[length(sel)]] - rtim[sel[1L]]) /
+                     max(1, (length(sel) - 1)))
                 maxi <- which.max(mtx[, 3L])
                 res[i, c("rt", "maxo")] <- mtx[maxi[1], c(1, 3)]
                 res[i, c("rtmin", "rtmax")] <- rtr
--- a/R/functions-XCMSnExp.R
+++ b/R/functions-XCMSnExp.R
@@ -283,19 +283,17 @@ dropGenericProcessHistory <- function(x, fun) {
         if (length(mtx)) {
             ## mtx: time, mz, intensity
             if (any(!is.na(mtx[, 3]))) {
-                ## How to calculate the area: (1)sum of all intensities / (2)by
-                ## the number of data points (REAL ones, considering also NAs)
-                ## and multiplied with the (3)rt width.
-                ## (1) sum(mtx[, 3], na.rm = TRUE)
-                ## (2) sum(rtim >= rtr[1] & rtim <= rtr[2]) - 1 ; if we used
-                ## nrow(mtx) here, which would correspond to the non-NA
-                ## intensities within the rt range we don't get the same results
-                ## as e.g. centWave. Using max(1, ... to avoid getting Inf in
-                ## case the signal is based on a single data point.
-                ## (3) rtr[2] - rtr[1]
+                ## Area = sum(intensities) * rt_width, where rt_width is the mean
+                ## scan spacing = (actual rt span of the scans in the window) /
+                ## (number of those scans - 1). Using the *actual* data span
+                ## (rather than the requested window rtr[2] - rtr[1]) keeps the
+                ## filled 'into' consistent with centWave detection, which
+                ## derives pwid from the peak's actual scan times, and avoids
+                ## inflating 'into' when the feature window overhangs the
+                ## sample's local scan coverage. max(1, ...) guards single-scan.
+                rt_in <- rtim[rtim >= rtr[1] & rtim <= rtr[2]]
                 res[i, "into"] <- sum(mtx[, 3L], na.rm = TRUE) *
-                    ((rtr[2] - rtr[1]) /
-                     max(1, (sum(rtim >= rtr[1] & rtim <= rtr[2]) - 1)))
+                    (diff(range(rt_in)) / max(1, (length(rt_in) - 1)))
                 maxi <- which.max(mtx[, 3L])
                 res[i, c("rt", "maxo")] <- mtx[maxi[1], c(1, 3)]
                 res[i, c("rtmin", "rtmax")] <- rtr
--- a/tests/testthat/test_XcmsExperiment-functions.R
+++ b/tests/testthat/test_XcmsExperiment-functions.R
@@ -314,3 +314,20 @@ test_that(".xmse_combine works", {
 
     expect_error(.xmse_combine(list(a, 3)), "objects accepted")
 })
+
+test_that(".chrom_peak_intensity_centWave into is invariant to window padding", {
+    mkpm <- function(mz, i) cbind(mz = mz, intensity = i)
+    rt <- c(10, 20, 30, 40)
+    x  <- list(mkpm(c(100, 100.1), c(50, 60)), mkpm(c(100, 100.1), c(900, 800)),
+               mkpm(c(100, 100.1), c(40, 30)), mkpm(c(100, 100.1), c(10, 5)))
+    cn <- c("mz", "mzmin", "mzmax", "rt", "rtmin", "rtmax", "into", "maxo",
+            "sn", "sample")
+    pa <- function(rtmin, rtmax) matrix(
+        c(100, 99.9, 100.2, (rtmin + rtmax) / 2, rtmin, rtmax), nrow = 1,
+        dimnames = list("F", c("mz", "mzmin", "mzmax", "rt", "rtmin", "rtmax")))
+    into <- function(peakArea) xcms:::.chrom_peak_intensity_centWave(
+        x, rt, peakArea, cn = cn, sampleIndex = 1L)[, "into"]
+    ## The same 4 scans are selected by both windows; a wider *requested*
+    ## window must NOT inflate `into` (rt_width uses the actual scan span).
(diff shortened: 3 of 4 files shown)


R/functions-XCMSnExp.R
```diff
@@ -283,19 +283,17 @@ dropGenericProcessHistory <- function(x, fun) {
...
-                ## (3) rtr[2] - rtr[1]
+                ## Area = sum(intensities) * rt_width, where rt_width is the mean
+                ## scan spacing = (actual rt span of the scans in the window) /
+                ## (number of those scans - 1). Using the *actual* data span
+                ## (rather than the requested window rtr[2] - rtr[1]) keeps the
+                ## filled 'into' consistent with centWave detection, which
+                ## derives pwid from the peak's actual scan times, and avoids
+                ## inflating 'into' when the feature window overhangs the
+                ## sample's local scan coverage. max(1, ...) guards single-scan.
+                rt_in <- rtim[rtim >= rtr[1] & rtim <= rtr[2]]
```
jorainer: this is not efficient. there's no need to subset retention times again. `mtx` contains the retention time, m/z and intensity values of the current subset.

R/functions-XCMSnExp.R
```diff
@@ -283,19 +283,17 @@ dropGenericProcessHistory <- function(x, fun) {
...
                 res[i, "into"] <- sum(mtx[, 3L], na.rm = TRUE) *
-                    ((rtr[2] - rtr[1]) /
-                     max(1, (sum(rtim >= rtr[1] & rtim <= rtr[2]) - 1)))
+                    (diff(range(rt_in)) / max(1, (length(rt_in) - 1)))
```
jorainer: here `(mtx[nrow(mtx), 1L] - mtx[1L, 1L]) / max(1, between(rtim, rtr) - 1)` would be more efficient.

R/XcmsExperiment-functions.R
```diff
@@ -557,7 +557,7 @@
                 res[i, c("rt", "mz", "maxo", "into")] <- c(
                     rts[maxi], mmz, mat[maxi, 2L],
                     sum(mat[, 2L], na.rm = TRUE) *
-                    ((rtr[2L] - rtr[1L]) / max(1L, (length(keep) - 1L)))
+                    (diff(range(rt[keep])) / max(1L, (length(keep) - 1L)))
```
jorainer: also here, why a `diff()` `range()`, `rts` is a `numeric` with the retention times, ordered by retention time. So, `rts[length(rts)] - rts[1L]` should be used in stead.

jorainer (changes requested):
Thanks Jan for the PR. But I have mixed feelings on the Claude contribution. the code was not ideal and the use case/example looks pretty artificial.

stanstrup:
Hi, thanks for taking a look. I tried to fix the code.
I admit the cases where this matters is limited. I just didn't like the idea that peak picking and gap-filling would give different results.

Here is an example that shows when it can matter.
worst_gapfill_plot

Btw.. I noticed that the raw data is not completely cut the same way as the processed data which caused me some confusion when digging into this.

```r
library(dplyr)
library(Spectra)
library(xcms)

x <- loadXcmsData("xmse")
temp <- filterFile(x, file = 1)
range(rtime(temp))/60
# [1] 42.62884 71.11719


files <- fileNames(temp) %>% gsub("/usr/local/lib/R/host-site-library", .libPaths()[1], .)


Spectra(files) %>% rtime %>% range %>% {./60}
# [1] 41.68963 74.99707
```

jorainer:
could be that `filterFile` drops the adjusted retention times if you subset to a file and that's why you see differences? note that also alignment can have an influence on the gap-filling/integration: peak detection is done **before**  alignment, the gap-filling **after**, so that might influence the calculation and an into integrated by peak detection might not be identical to the one at gap-filling (because rt-s can change)... just to note.

otherwise I agree, better to stick with the original (or same) definition.

stanstrup:
Good idea. rtime() does indeed return the adjusted time so not comparable to the raw data. But that is not the reason here. They seem to be cut differently also looking at the scans. Doesn't matter for this PR it was just a curiosity I noticed. They are both cropped; but slightly differently.

```r
library(dplyr)
library(Spectra)
library(xcms)

## included dataset
x <- loadXcmsData("xmse")
temp <- filterFile(x, file = 1)
range(rtime(temp))/60
# [1] 42.62884 71.11719

range(spectraData(spectra(temp))[, "rtime"])/60
# [1] 42.52428 70.82385

range(spectraData(spectra(temp))[, "rtime_adjusted"])/60
# [1] 42.62884 71.11719

head(spectraData(spectra(temp))[, "spectrumId"])
# [1] "scan=33" "scan=34" "scan=35" "scan=36" "scan=37" "scan=38"

tail(spectraData(spectra(temp))[, "spectrumId"])
# [1] "scan=1113" "scan=1114" "scan=1115" "scan=1116" "scan=1117" "scan=1118"

## Original file
files <- fileNames(temp) %>% gsub("/usr/local/lib/R/host-site-library", .libPaths()[1], .)

Spectra(files) %>% rtime %>% range %>% {./60}
# [1] 41.68963 74.99707

head(spectraData(Spectra(files))[, "spectrumId"])
[1] "scan=1" "scan=2" "scan=3" "scan=4" "scan=5" "scan=6"

tail(spectraData(Spectra(files))[, "spectrumId"])
# [1] "scan=1273" "scan=1274" "scan=1275" "scan=1276" "scan=1277" "scan=1278"
```

R/functions-XCMSnExp.R
```diff
@@ -283,19 +283,20 @@ dropGenericProcessHistory <- function(x, fun) {
...
+                ## centWave detection, which derives pwid from the peak's actual
+                ## scan times, and avoids inflating 'into' when the feature
+                ## window overhangs the sample's local scan coverage. Numerator
+                ## and denominator use the same scan set (all scans in the
+                ## window); max(1, ...) guards the single-scan case.
+                sel <- which(between(rtim, rtr))
                 res[i, "into"] <- sum(mtx[, 3L], na.rm = TRUE) *
-                    ((rtr[2] - rtr[1]) /
-                     max(1, (sum(rtim >= rtr[1] & rtim <= rtr[2]) - 1)))
+                    ((rtim[sel[length(sel)]] - rtim[sel[1L]]) /
```
jorainer, line 298: please, no, make manual changes no Claude. this is **not** what I requested. `mtx` is a `matrix` that has the retention times of the peaks within the range in its first column. So, using `mtx[1L, 1L]` and `mtx[nrow(mtx), 1L]` will give you the minimum and maximum (actual) retention times of the signal within the area.

then, for the number of spectra acquired within the retention time range for this peak: `max(1, (sum(between(rtim, rtr)) - 1))`

so:

`(mtx[nrow(mtx), 1L] - mtx[1L, 1L]) / max(1, sum(between(rtim, rt)) - 1)` would be the version that needs less copying/functions.

jorainer (changes requested):
Please make changes as requested. And **don't** use Claude for that. I have to repeat asking for changes that Claude is then either ignoring or suggesting different solutions. This is wasting my time. I'm not at all happy with AI contributions considering their waste of energy (and of my time).

And please update the DESCRIPTION increasing the version number and add a NEWS entry.

stanstrup:
Thanks for the review. You're right about the efficiency. I replaced `diff(range())` with direct first/last indexing in both files.

I did not use your suggestion literally because I do not think it is the right fix here. I am sorry I didn't explain the difference before.
In your version the number of scans cover the whole range, but the RT only covers from the first to the last non-zero scan. That occasionally makes a difference.

I can show you here. Need a bit of setup to walk into the right place in the code so forgive me for the amount of code copied from the functions




## Setup

```r
library(xcms)
library(BiocParallel)
library(MsCoreUtils)
library(Spectra)
register(SerialParam())

x0 <- dropFilledChromPeaks(loadXcmsData("xmse"))

## Make internal functions callable
.features_ms_region <- xcms:::.features_ms_region
.pmat_filter_mz     <- xcms:::.pmat_filter_mz
```


## Build the inputs to `.chrom_peak_intensity_centWave`

What `fillChromPeaks(ChromPeakAreaParam())` does before calling the
integrator (lines 1971–2011 in XcmsExperiment.R):

```r
param <- ChromPeakAreaParam()
feature_ids <- rownames(featureDefinitions(x0))
fr <- .features_ms_region(
    x0, mzmin = param@mzmin, mzmax = param@mzmax,
    rtmin = param@rtmin, rtmax = param@rtmax,
    minMzWidthPpm = param@minMzWidthPpm,
    features = feature_ids)
fr <- cbind(fr, mzmed = featureDefinitions(x0)$mzmed)
fvals <- featureValues(x0, value = "index")
pal <- lapply(seq_len(ncol(fvals)), function(i) {
    fr[is.na(fvals[, i]), , drop = FALSE]
})
```

Then `.xmse_integrate_chrom_peaks` (lines 500–513) extracts per-sample
`rt` and `x` and calls `.chrom_peak_intensity_centWave`:

```r
sp <- spectra(x0)
keep_ms1 <- which(sp$msLevel == 1L)
f <- as.factor(fromFile(x0)[keep_ms1])
rt_all <- sp$rtime_adjusted[keep_ms1]
cn <- colnames(chromPeaks(x0))

pd <- peaksData(filterMsLevel(sp, 1L), f = factor(),
                return.type = "list", BPPARAM = SerialParam())
pd_s <- split(pd, f)
rt_s <- split(unname(rt_all), f)
```


## Walk through `.chrom_peak_intensity_centWave`

Two examples that show the difference.

### Case 1: all signal in one scan (sample 2, area 71)

```r
s        <- 2L
x        <- pd_s[[as.character(s)]]
rt       <- rt_s[[as.character(s)]]
peakArea <- pal[[s]]
```

Lines 541–553: filter to the peak area and build `rts`:

```r
i    <- 71L
rtr  <- peakArea[i, c("rtmin", "rtmax")]
keep <- which(between(rt, rtr))              # 14 scans in window
xsub <- lapply(x[keep], .pmat_filter_mz,
               mzr = peakArea[i, c("mzmin", "mzmax")])
mat  <- do.call(rbind, xsub)                 # 1 row — one scan has signal
nr   <- vapply(xsub, nrow, NA_integer_)
rts  <- rep(rt[keep], nr)                    # length 1 — only signal scans
```

14 scans fall inside `rtr`, but only 1 carries signal at this m/z.
So `rts` has 1 entry, not 14 (see explanation below case 2).

**Line 559 — the one we are discussing:**

```r
S <- sum(mat[, 2L], na.rm = TRUE)                    # 17,632

## Committed (rt[keep] — all scans in the window):
into <- S * ((rt[keep[length(keep)]] - rt[keep[1L]]) /
                 max(1L, length(keep) - 1L))         # 27,157

## Suggested (rts — only scans with signal):
into <- S * ((rts[length(rts)] - rts[1L]) /
                 max(1L, length(keep) - 1L))         # 0
```

| | time span | scan count − 1 | spacing | into | measured spacing |
|---|---:|---:|---:|---:|---:|
| **`rt[keep]`** (all window scans) | 20.02 s | 13 | 1.540 s | 27,157 | 1.541 s |
| **`rts`** (signal scans only) | 0.00 s | 13 | 0.000 s | **0** | 1.541 s |

`rts` has one entry, so `rts[1] − rts[1] = 0` → `into = 0` while
`maxo = 17,632`.


### Case 2: signal covers part of the window (sample 3, area 100)

```r
s        <- 3L
x        <- pd_s[[as.character(s)]]
rt       <- rt_s[[as.character(s)]]
peakArea <- pal[[s]]
```

Lines 541–553 (same as above, for sample 3, i = 100):

```r
i    <- 100L
rtr  <- peakArea[i, c("rtmin", "rtmax")]
keep <- which(between(rt, rtr))              # 38 scans in window
xsub <- lapply(x[keep], .pmat_filter_mz,
               mzr = peakArea[i, c("mzmin", "mzmax")])
mat  <- do.call(rbind, xsub)
nr   <- vapply(xsub, nrow, NA_integer_)
rts  <- rep(rt[keep], nr)                    # 22 entries — signal scans
```

38 scans in window, 22 carry signal.

| | time span | scan count − 1 | spacing | into | measured spacing |
|---|---:|---:|---:|---:|---:|
| **`rt[keep]`** (all window scans) | 58.47 s | 37 | 1.580 s | 14,474,443 | 1.578 s |
| **`rts`** (signal scans only) | 55.29 s | 37 | 1.494 s | 13,685,533 | 1.578 s |

 `rts` only starts at the first scan with signal and ends at the
last scan with signal, so the span misses the empty scans at the
edges of the window. The numerator is too short for the denominator.


`rts` is built as `rep(rt[keep], nr)` where `nr` counts the rows
returned by `.pmat_filter_mz` per scan. Scans that fall inside `rtr`
but have no peak at `mzr` get `nr = 0` and produce no entry in `rts`.
The denominator `length(keep) - 1` counts all scans in the window.

--> using `rts` for the span and `keep` for the scan count mixed different ranges.

The same applies to `.getChromPeakData`: `mtx` (from `.rawMat`) only
contains rows where the scan has signal at the requested m/z, so
`mtx[nrow(mtx), 1L] - mtx[1L, 1L]` is the signal-scan span, not the
window-scan span.


## It happens occasionally

```r
## XcmsExperiment path (loadXcmsData):   5 of 975 filled peaks -> into = 0 with rts
## XCMSnExp path (faahKO, 3 files):      7 of 186 filled peaks -> into = 0 with mtx
## rt[keep] / rtim[sel] gives into = 0 for none.
```

You could argue that the area of a zero-length peak should be zero. But I'd argue that for our purpose it is better to consider it 1 scan long.


## Plot

Scan spacing comparison


Top row: all signal in one scan. `rts` span = 0, so `into = 0` while
`maxo` is nonzero.

Bottom row: signal covers part of the window. Both give nonzero
values, but only `rt[keep]` (green) lands on the measured scan spacing
(dashed line).



For DESCRIPTION and NEWS: #841 changed both, so I am not sure what to do here to avoid merge conflicts.

PR #846, opened 2026-08-24
fix: workaround for error reading character subsets from HDF5

jorainer:
- Reading subsets of a character array defined by indices from an HDF5 file causes errors. The bug is reported for *rhdf5* in issue https://github.com/Huber-group-EMBL/rhdf5/issues/240 . This commit adds a workaround that reads the full character vector and subsets it by index in R.


sneumann (approved):
Hi, looks great, passes check here locally as well. Will merge and push. Yours, Steffen

